PR #37, opened 2019-04-05
added what management should do

### LINKED ISSUE #21: Adhere to Open Standards: add field Management, What You Need To Do

nebster:
fixes #21

### DIFF
--- a/standards/open-standards.md
+++ b/standards/open-standards.md
@@ -33,7 +33,7 @@ order: 10
 
 ## Management: what you need to do
 
-* ???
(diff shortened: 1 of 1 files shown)


standards/open-standards.md
```diff
@@ -33,7 +33,7 @@ order: 10
 
 ## Management: what you need to do
 
-* ???
+* MAY Include open standard compliance assessment in code reviews.
```
bvhme: Everywhere else we're not using these codewords outside of the 'Measure', but this seems fine

PR #54, opened 2019-04-09
Fixes issues 2 and 16 for standards 10-14.

ElenaFdR:
Some language tidying as well.


standards/bundle-policy-and-code.md
```diff
@@ -18,12 +18,19 @@ order: 2
 
 ## What this doesn’t do
 
+* This process does not guarantee that a codebase will reflect the bundled policy.
```
bvhme: Perhaps it should... But that might be a new issue.

PR #79, opened 2019-04-12
Fixes #30 (how to test)

### LINKED ISSUE #30: Style: standardise statements under 'how to test'
Suggestion: frame these as either positive statements or questions (not a hybrid).
Current example from Code in the open:
"Is the current version of the source published on the internet where it can be seen from outside your organization and without the need for any form of authentication."
Suggested alternatives:
- Is the current version of the source published on the internet where it can be seen from outside your organization and without the need for any form of authentication?
- The current version of the source is published on the internet where it can be seen from outside your organization and without the need for any form of authentication.

ElenaFdR:


standards/documenting.md
```diff
@@ -30,13 +30,14 @@ order: 8
 
 ## How to test
 
-* test readability of documentation with other stakeholders, professionals from other public organisations and the general public
-* generate documentation from code
-* test links and images
+* Other stakeholders, professionals from other public organisations and the general public find the documentation clear and understandable
```
bvhme: I feel this one is a bit hard to continously test, is there a way we could store this information so that it can be referenced at every check, otherwise I feel we might need to solve this differently

PR #81, opened 2019-04-14
changed 3rd sentence of 'Measure', added '.', n't>not, prioritise>pri…

MirjamvT:
…oritize

not sure if the doc is to be in UK or US English, think i saw some more US English, hence the change of prioritize into prioritize.


MirjamvT:
Hi
i understand your point.
should we then specify 'the organization'? it is unclear to me which
organization is meant (of the coder? or of the org 'requesting the
codebase'? or of the organizations covered by the encoded policy?
this is why i contracted organization with third parties, inside and
outside. as it seems to pertain to all of the above?

Best regards,
Mirjam van Tiel
+31 6 4662 8454

ElenaFdR:
I'm closing this now - all of its contents have been acted on already (one way or another).

PR #83, opened 2019-04-14
n't > not

MirjamvT:


standards/bundle-policy-and-code.md
```diff
@@ -2,7 +2,7 @@
 order: 2
 ---
 
-# Bundle policy and source code
+# For each codebase: bundle policy and source code
```
bvhme: I think this is not an improvement per sé, although the extra context is useful it is too far out of line with other standards I feel.

I suggest reverting to the original.

PR #90, opened 2019-04-15
Improved naming for sections

### LINKED ISSUE #44: Style: standardise what we call the 15 points/criteria in the standard.
I suggest 'points' for simplicity (GDS does this - see https://www.gov.uk/service-manual/service-standard/understand-user-needs).
OSI call them 'criteria' (see https://opensource.org/osr)
Both pass the Google Translate test.

And then start using this in all of our discussion of the standard, so that we consistently set people's expectations.

bvhme:
* Standards are now named _criteria_
* The measure is now named _requirements_
* 'What this doesn't do' is now 'what this does not do' 

Fixes #44
Closes #85 Closes #83


nebster:
I think Criteria is not very accessible.  Should we not go with Principle, Pattern, or Point?

bvhme:
I suggest merging this now and solving that issue later

PR #94, opened 2019-04-15
used policy > policy in use

MirjamvT:
I assume we mean to talk about policy and code that is in use at any given time? see also the sentence in next quoted bullit.
(or are we looking for a term meaning ‘geldig’? but ‘valid’ seems odd).

wondering: used by whom?


ElenaFdR:
Re in use: it's in opposition to the next sentence (about code not currently in use, which is better defined).

PR #97, opened 2019-04-15
make this sentence an 'active one' too

MirjamvT:


bvhme:
The test here is to make sure that a version is published on the internet, not to publish it. It is an outcome statement that tests for 'if you look at where it should be it is there'

MirjamvT:
this 'how to test' header (not a question, though in question form, grammatically) is still puzzling to me. it does not seem to intuitively match the sentence below.
how about turning the text below into a question ('is the code...'), a 'yes' answer would then be a 'pass' on the test.

bvhme:
I think that is a great idea, however that should be an issue and not a Pull Request :)

Would you like to add it or shall I?

MirjamvT:
After reading the intro (wasn't avaible at the time i made this PR i think) I understand 'How to test' to mean 'How to test if a contribution is compliant with this standard'.
will make a pull request for making the header readable 'on its own' (and longer, regrettably)

PR #98, opened 2019-04-15
added policy to source code as underpinning, and:

MirjamvT:
re underpinning of interactions b/w 'general public' and 'your organization'
ALSO: should we write 'users' instead of 'general public'
ALSO: should we write 'a contributing public administration' instead of 'your organization' (since we are writing text in 3rd person singular)


bvhme:
I am afraid 'users', 'general public' etc mean very specific things. We should care to take these into our glossary.

Users are whomever uses the software (which could well be the organizations themselves).

PR #100, opened 2019-04-15
made the headers more specific

MirjamvT:
by adding the title of the measure to the headers, the reader cannot be confused what the header refers to. a reader might think 'this' (old wording) is the standard, use of the standard or such. but we are talking about the particular measure here. curious what you think.


bvhme:
I feel that because every standard has these same points it is actually easier to follow if they are the same.

Closing for now.

PR #112, opened 2019-04-15
added 'code' to write 'source code'.

MirjamvT:


bvhme:
Perhaps it is worth just naming it 'Source' and clarifying in the glossary what that means in order to make the difference with policy clear.

PR #121, opened 2019-04-16
Improve introduction

bvhme:
I've reorganised the introduction to make a bit more sense following Martijn's feedback as well as removing or correcting some problems.

I've also tried to make it a bit more of a general problem solving document as opposed to something purely for our use.

### DIFF
--- a/introduction.md
+++ b/introduction.md
@@ -2,7 +2,7 @@
 
 Public code is both civil code (like policy or regulation) and computer source code (such as software and algorithms) executed in a public context, by humans or machines.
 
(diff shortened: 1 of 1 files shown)


ElenaFdR:
I've made some sentences clearer. I don't think the certification process section works at the moment - at a minimum, it relies on information that's currently placed beneath it, like the goals for the Standard. In the future, I suggest putting all info on certification on a separate page hosted on about.publiccode.net - I'd regard how to get code certified as being integral to the working of the Foundation, but not integral to understanding and applying the Standard to your work.

PR #122, opened 2019-04-16
took out 'current version'

MirjamvT:
This ‘current version’ phrase confuses me as there seems to be an ever-changing version, because anyone can make changes at any time, forever. Maybe better to write something like: We publish the source on the internet where… Or even: We publish the ever-developing source…
and.. i snuck back in the word 'code' after 'source'.


bvhme:
This is meant to indicate whatever version is currently published

PR #130, opened 2019-04-16
public policy and reduce requirements

MirjamvT:
also: 
does the 3rd requirement make the 1rst and 2nd redundant?
('All policy and source code that the codebase is based upon MUST be documented, reusable and portable' seems to ask for inclusion of policy and code?)


bvhme:
It doesn't, and means something different, also we call it _policy_ everywhere else in the standard, why would we add _public_ specifically here?

PR #136, opened 2019-04-18
new lemma:  continuous integration

MirjamvT:
def needs to be improved if this lemma stays in


ElenaFdR:
Closing because continuous integration is well explained on the internet - we don't need to provide our own definition. We already link to 2 'what is continuous delivery' articles in 'Further reading' on 'Use continuous delivery'.

PR #138, opened 2019-04-18
clarify the 'test'- header

MirjamvT:


bvhme:
As we've discussed this should be an issue, and submitted into one large pull request changing all the headers as one. Closing this one for now.

@MirjamvT will you take care that there is a good issue describing this?

PR #163, opened 2019-04-18
Issue with causality in sentence

clausmullie:
I feel documenting does not lead to help others to ask this question.
It does however encourage those who already have asked this question.

### DIFF
--- a/criteria/documenting.md
+++ b/criteria/documenting.md
@@ -21,7 +21,7 @@ order: 8
 ## What this does
 
 * Users can start using and contributing more quickly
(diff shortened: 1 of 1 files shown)


ElenaFdR:
Clear documentation should help with SEO and people's ambient knowledge (so that we're not only reaching people trying to explicitly solve that problem at that time.)

Proposed alternative text:
* Helps people discover your codebase, especially people asking 'is there already code that does something like this?'

clausmullie:
@ElenaFdR fixed!

ElenaFdR:
@clausmullie Ready to merge!

PR #201, opened 2019-05-13
use reusable instead of useable

MirjamvT:
for more consistent use of the term reusable

### DIFF
--- a/introduction.md
+++ b/introduction.md
@@ -59,7 +59,7 @@ Most proprietary software systems currently used by public organizations do not 
 We consider public code to have these core values:
 
 * Inclusive
-* Usable
+* reusable
 * Open
 * Legible
 * Accountable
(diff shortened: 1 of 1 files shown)


introduction.md
```diff
@@ -59,7 +59,7 @@ Most proprietary software systems currently used by public organizations do not
 We consider public code to have these core values:
 
 * Inclusive
-* Usable
+* reusable
```
bvhme: What about a capital here :)

```suggestion
* Reusable
```

bvhme:
@nebster, @clausmullie and @ElenaFdR, I'm wondering what you are thinking about this, as it is quite a core change.

ElenaFdR:
Is it worth putting this as an issue, rather than a PR? This also ties into #203 and #208.

bvhme:
For now I think this is conflating two goals, namely: to make software that is easy to use and to make software that can be redeployed by different organisations.

I'm going to close this, but we have to be aware these two realities are existant.

PR #220, opened 2019-06-12
Minor updates and additions to different chapters.

petkivim:
* Open licenses
  * License header should be included in source code files
* Open to contributions
  * The codebase should have a contributor license agreement
* Version control and history
  * Use tags for marking different versions of the codebase

### DIFF
--- a/criteria/advertise-maturity.md
+++ b/criteria/advertise-maturity.md
@@ -12,7 +12,7 @@ order: 14
     * alpha - to do guided tests with a limited set of users
     * beta - to open up testing to a larger section of the general public, for example to test if the codebase works at scale
     * pre-release version - code that is ready to be released but hasn't received formal approval yet
-* A codebase SHOULD contain a log of changes from version to version
+* A codebase SHOULD contain a log of changes from version to version, for example in the `CHANGELOG`
 
 ## Why this is important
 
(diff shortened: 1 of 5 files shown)


bvhme:
Thanks for the awesome PR @petkivim,

I think there are a few things in there we might want to foster a discussion on before merging this. We'll keep you updated!

ericherman:
Hello @petkivim,

I apologize for the overly long delay.

There are several topics in this pull request and I'd like to ask if you are willing to break this PR up into several very small PRs. For instance, "clear commit messages" has little to do with "passwords in production systems" or "copyright/license headers". Restructuring as several tiny PRs would allow us to merge some of your contributions right away.

Regarding use of tags, the standard is not technology specific, and this seems to me to be a bit too git-centric. Please consider if there is a revision-control-system-agnostic way of wording of this, perhaps with a "for instance" which is git-specific.

The Reuse Specification isn't something that we're specifically recommending at this time. A separate PR on this topic would be a good place to kick-off a conversation about this topic. Maybe the result of that conversation would be to link to the Reuse Specification as a "for instance" or maybe the Foundation would adopt it as a part of the Standard. Also the link resolves to version 3 of the spec.

There is an existing discussion on CLAs: https://github.com/publiccodenet/standard/issues/211 .... Perhaps you have additional thoughts to add to that discussion? Regardless, adding CLAs would need to be a separate PR and should connect to that issue.

Thanks for your contributions,
 -Eric

petkivim:
Hi @ericherman 

Thank you for the feedback. I splitted the PR into multiple small PRs according to your suggestions. This PR can be closed now.

Thanks,
Petteri

Issue #205, opened 2019-05-13
Add easier/manual ways of providing feedback to the contributor guide


bvhme:
Reading the printed version @EdoPlantinga was overwhelmed by the complexity (especially for non-nerds). Perhaps we can mention emailing feedback before diving into how to make changes on GitHub.

ElenaFdR:
I've added a sentence offering <info@publiccode.net>, but it might be more useful to create <standard@publiccode.net>.

bvhme:
Maybe <standard-feedback@publiccode.net> would be easier to use. I'm guessing this should be a group, and who needs to be in it?

ElenaFdR:
👋

ericherman:
We have mailing list:
* standard@lists.publiccode.net

* https://lists.publiccode.net/mailman/postorius/lists/standard.lists.publiccode.net/

And soon it will be in the contributing guide: #422

ericherman:
Resolved by #503

Issue #204, opened 2019-05-13
Glossary in the beginning


bvhme:
It might make sense to explain the key terms in this booklet before the reader gets knee-deep in our own invented terminology. Perhaps after the readers guide and before the criteria.

ElenaFdR:
Do we need to update how the PDF is generated as well?

ElenaFdR:
Nevermind - I've updated print.html now as well.

Issue #197, opened 2019-05-09
Bullet sounds skeptical about policymakers' ability to understand code


ElenaFdR:
This bullet in [bundle policy and source code](http://standard.publiccode.net/criteria/bundle-policy-and-code.html) is easy to read in a deeply sarcastic tone of voice:

```
* Try to periodically understand what the non-policy code in the codebase does
```
Suggested alternative:
```
* Check in regularly to understand how the non-policy code in the codebase has changed and what it's doing now

ElenaFdR:
@bvhme Any thoughts?

bvhme:
I agree, let me do another suggestion:

> * Check in regularly to understand how the non-policy code in the codebase has changed and whether that is still in line with the intentions of the policy.

ericherman:
(merged)

PR #227, opened 2019-06-19
Style proofread

ElenaFdR:
This should close:
- #197 
- #204
- #205 

(I'll update this list as I work through this)

### DIFF
--- a/criteria/open-licenses.md
+++ b/criteria/open-licenses.md
@@ -7,7 +7,7 @@ order: 12
 ## Requirements
 
 * All code and documentation MUST be published under a license that allows it to be freely reusable, changeable and redistributable
-* Software source code this MUST be licensed under an [OSI-approved open source license](https://opensource.org/licenses/category)
+* Software source code MUST be licensed under an [OSI-approved open source license](https://opensource.org/licenses/category)
 * All code MUST be published with a license
 * Codebases MAY have multiple licenses for different types of code and documentation
 * Documentation MAY be published under [Creative Commons licenses
--- a/criteria/open-licenses.md
+++ b/criteria/open-licenses.md
@@ -8,7 +8,6 @@ order: 12
 
 * All code and documentation MUST be published under a license that allows it to be freely reusable, changeable and redistributable
 * Software source code MUST be licensed under an [OSI-approved open source license](https://opensource.org/licenses/category)
-* All code MUST be published with a license
 * Codebases MAY have multiple licenses for different types of code and documentation
 * Documentation MAY be published under [Creative Commons licenses
 ](https://creativecommons.org/licenses/) that are NOT 'no derivatives' or 'non-commercial'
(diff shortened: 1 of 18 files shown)


criteria/open-licenses.md
```diff
@@ -7,7 +7,7 @@ order: 12
 ## Requirements
 
 * All code and documentation MUST be published under a license that allows it to be freely reusable, changeable and redistributable
-* Software source code this MUST be licensed under an [OSI-approved open source license](https://opensource.org/licenses/category)
+* Software source code MUST be licensed under an [OSI-approved open source license](https://opensource.org/licenses/category)
 * All code MUST be published with a license
```
ElenaFdR: This appears to be the same as the first bullet, just less specific. If so, can we delete this? If not, can we make this more clear?

ElenaFdR: If it's meaningfully different from the first bullet, than I suggest moving this to the top of the list so that it goes from broader to narrower as you go down.

criteria/open-standards.md
```diff
@@ -9,9 +9,9 @@ order: 10
 * For features of a codebase that facilitate the exchange of data the codebase MUST use an open standard that meets the [Open Source Initiative Open Standard Requirements](https://opensource.org/osr)
 * If no existing open standard is available, effort SHOULD be put into developing one
 * Standards that are machine testable SHOULD be preferred over those that are not
-* Functionality using features from a non-open standard (one that doesn't meet the [Open Source Initiative Open Standard Requirements](https://opensource.org/osr)) MAY be provided if necessary, but only in addition to compliant features.
+* Functionality using features from a non-open standard (one that doesn't meet the [Open Source Initiative Open Standard Requirements](https://opensource.org/osr)) MAY be provided if necessary, but only in addition to compliant features
 * All non-compliant standards used MUST be documented clearly in the documentation
-* The codebase MAY contain a list of all the standards used with links to where they are available
+* The codebase SHOULD contain a list of all the standards used with links to where they are available
```
ElenaFdR: It's significantly easier to test for this criterion if there's a list in the documentation. Is there any reason not to make this MUST?

criteria/open-licenses.md
bvhme: Great idea!

ElenaFdR: Which comment are you replying to?

bvhme: The second one :)

criteria/open-standards.md
bvhme: I've made new issue https://github.com/publiccodenet/standard/issues/238 for this so this doesn't block the merge.

criteria/version-control-and-history.md
```diff
@@ -17,17 +17,18 @@ order: 5
 
 Version control means keeping track of changes to your code over time. This allows you to create structured documentation of the history of the codebase. This is essential for collaboration at scale.
 
-Version control enables you to:
+Version control allows you to:
```
bvhme: This is a definite 'enable'. To allow is to give permission, to enable is to make possible.

Here is an article from Haje on this: https://medium.com/@Haje/in-tech-journalism-you-rarely-allow-you-probably-enable-8eb918a6af15 :)

ElenaFdR: Yup, I agree.

PR #251, opened 2019-08-16
Marking different versions in Version Control and History

petkivim:
- Accessing a specific version is easier when different versions are marked.
- Without marking it is more challenging to find the commit that was used as a base for a release.
- Marking is aligned with labels that are mentioned in the Advertise Maturity chapter (14).


ericherman:
It's been a while since I've used SVN, but that system uses path and name if I recall, which I think this wording more-or-less covers.

I see the "This branch has conflicts that must be resolved" note, so you might want to rebase, but otherwise looks fine.

PR #252, opened 2019-08-16
Add specification regarding the use of production data or secrects in…

petkivim:
… Create reusable and portable code

- Including fake data as well as test tokens and passwords in tests should be allowed.
- Code in the open chapter (1) also says that production data and secrets must not be included in contributions.


petkivim:
Merged changes from develop.

PR #257, opened 2019-08-20
Code is in use by multiple parties

ericherman:
Codebases without communinities are either too narrow for stewardship or
must continue under incubation until a self-sustaining community can be
built.

-----
[View rendered codebase-criteria/reusable-and-portable-codebases.md](https://github.com/publiccodenet/standard/blob/multi-party/codebase-criteria/reusable-and-portable-codebases.md)


bvhme:
With this addition we might need to be a bit clearer about what the path to mature means and how to get there.

PR #271, opened 2019-09-10
Fixed typo, added European Accessibility Directive and WCAG 2.1 on readability

ElenaFdR:
This is an incomplete PR because:

- there are probably tools to test the accessibility level (I could do more research on this)
- it's misleading/reductive to bundle web accessibility and related legal requirements under 'plain English'
- I'm not 100% confident how the EU accessibility directive applies to developer docs - more research is needed! (@Ainali @ericherman @bvhme, do you know more about this?)

-----
[View rendered criteria/understandable-english-first.md](https://github.com/publiccodenet/standard/blob/readability-is-an-accessibility-issue/criteria/understandable-english-first.md)

### DIFF
--- a/criteria/understandable-english-first.md
+++ b/criteria/understandable-english-first.md
@@ -10,12 +10,14 @@ order: 9
 * Any translation MUST be up to date with the English version and vice-versa.
 * There SHOULD be no acronyms, abbreviations, puns or legal/domain specific terms in the codebase without an explanation preceding it or a link to an explanation.
 * The name of the project or codebase SHOULD be descriptive and free from acronyms, abbreviations, puns or branding.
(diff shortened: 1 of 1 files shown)


ElenaFdR:
A quick scan of the UK guidance on meeting the EU Accessibilty Directive ([this](https://www.gov.uk/guidance/accessibility-requirements-for-public-sector-websites-and-apps) and [this](https://www.gov.uk/guidance/make-your-website-or-app-accessible-and-publish-an-accessibility-statement)) suggests that any public facing public sector website or app needs to be compliant - so in the first instance, perhaps this only applies to codebases that either have a public front end or where there's a public comms website created for the codebase?

Ainali:
Yes, it is my understanding as well that the directive only applies to front ends that can be accessed by the general public (that doesn't mean it is not good practise to be WCAG compliant for other UI parts as well). Now I don't know exactly how this will affect a public body when publishing on a third party platform (like GitHub) where they might not have full control over all the things the directive covers. That being said, I think your suggestions are good and solid improvements.

PR #278, opened 2019-10-04
Minor editorial tweaks

skybristol:
I made a few editorial suggestions in the files of the standard as I was reading through (sorry, I couldn't help myself).

I really like what you all are doing with this standard. In my day job, I work in the federal sector in the US as a scientist in a non-regulatory context. My job is to study and provide information on what we observe and understand, but I'm very interested in the process of opening up the policymaking space to direct democratic involvement by the people whose lives are effected by government policies.

I was curious about one thing in your underlying philosophy. You state explicitly in your introduction, "It is not aimed at public organizations’ end users (residents or citizens)." I get where you're coming from to a certain extent, but I'm interested in your thoughts on how public policy software processes that help guide direction in our cities and regions may be contributed to directly by interested citizens. It seems like this is being done in various ways around the world through city hack-a-thons, code challenges, and less structured experiments. You have a great structure to the standard with very clear language targeted at your stated audiences, which I found quite compelling. But I'm curious about why the decision to explicitly not think about the implications to residents or citizens themselves.

-----
[View rendered introduction.md](https://github.com/skybristol/standard/blob/editorial-finds/introduction.md)
[View rendered readers-guide.md](https://github.com/skybristol/standard/blob/editorial-finds/readers-guide.md)

### DIFF
--- a/introduction.md
+++ b/introduction.md
@@ -31,14 +31,14 @@ Current public software production methods have not served public service delive
 In the last decade, public organizations that purchased complete software solutions have sometimes been surprised to discover that they:
 
 * can’t change their software to reflect changing policy or take advantage of new technology
-* don’t have access to their data as it's locked in to systems
+* don’t have access to their data as it's locked into proprietary systems
 * are asked to pay ever increasing license fees
 
 ### Technological sovereignty and democratic accountability
 
 Public institutions, civil servants and residents deserve better.
 
-We believe the software that runs our society can no longer be a black box, ordered from outside companies that keep the underlying logic on which their software operates hidden in proprietary codebases. Instead, governments need technological sovereignty - allowing them to set and control the functioning of public software, just like they are able to set and control policy that is legally formulated in laws. Citizens and civil society actors need this software to be transparent and accountable. The design of software as essential civic infrastructure should honor digital citizens’ rights.
+We believe the software that runs our society can no longer be a black box, controlled by outside companies that keep the underlying logic on which their software operates hidden in proprietary codebases. Instead, governments and the people they serve need technological sovereignty - allowing them to set and control the functioning of public software, just like they are able to set and control policy that is legally formulated in laws. Citizens and civil society actors need this software to be transparent and accountable. The design of software as essential civic infrastructure should honor digital citizens’ rights.
 
 ### Designing truly public software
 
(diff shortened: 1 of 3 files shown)


introduction.md
```diff
@@ -110,16 +110,16 @@ The audit is presented as a review of the contribution. The codebase steward giv
...
-If the audit process is added to an existing codebase, the new merge requests can be certified, but the existing code cannot be certified. By auditing every new merge request the codebase can move towards being completely certified.
+If the audit process is added to an existing codebase, the new merge requests can be certified, but the existing code cannot be certified. By auditing every new merge request the codebase can move incrementally towards being completely certified.
 
 ## The goals for the Standard for Public Code
 
 This Standard supports developers, designers, business management and policy makers to:
 
 * develop high quality software and policy for better public service delivery
-* develop reusable codebases that can be reused across contexts and collaboratively maintained
+* develop buildable codebases that can be reused across contexts and collaboratively maintained
```
ElenaFdR:
```suggestion
* develop codebases that can be reused across contexts and collaboratively maintained
```

I approve of the deduping of 'reusable', but 'buildable' doesn't seem directly relevant. Do you mean something closer to 'hackable' (as opposed to a monolithic black box)?

skybristol: Thanks for considering my suggestions. I concur with your suggested change.

It's not really a proper English word yet, but "buildability" is a different thing than reusability and maybe more powerful in the context of the public code standard you are promoting. Reusable connotes more a process of forking the code and using it in a completely different context, which is also cool. Buildable to me is more about either building onto a codebase in its original context, making it better, or using that codebase directly to build some other capability. It also relates to being able to complete a build process, technically in a CI/CD sense. In the same bullet, we have "collaboratively maintained," which is hugely important. But I'd also say maintenance implies potentially just keeping up with bug fixes and dependency changes, while if the core concept is buildable, that suggests that future collaboration might be adding features and capabilities.

Hackable could work in some respects but seems to connote a less formal process than buildable. It is also a good thing, though, because it implies a level of rapid accessibility and ease of experimentation, which I think we also want to promote.

All three of these are important concepts with slightly different semantics that are already somewhat addressed elsewhere in the criteria. For the intro, your suggested change makes good sense without confusing the situation with all this other stuff. I'll ponder a bit more on the semantics of reusable, buildable, and hackable and see if I come up with any other suggestions for elsewhere.

skybristol:
I don't mind being included as a contributor. This was a personal contribution and not one made in my official capacity as a government scientist.

I'd also value an opportunity to contribute something of more substance. I really am curious about your reasoning on explicitly not directing the Public Code Standard at residents/citizens. I'm sure you have some interesting thinking on that subject and would value a dialog. I can imagine areas of the Standard where I could articulate what the implications would be for residents of a municipality or other jurisdiction with interest in contributing directly to public code. But there are perhaps some really good arguments for not going there, and I'd love to hear more.

In my official capacity, I have pointed my team and some of my other colleagues at your work here and asked them to think about implications on our side of the pond and to the decision analysis/support work that we do at multiple levels of government. I look forward to following this work and seeing how we can plug in.

bvhme:
Thanks @skybristol, really valuable!

PR #288, opened 2019-11-06
Simplify & avoid 'code and docs under "a" licnese'

ericherman:
While it is implied that code and docs are not required to share a
license, the previous wording could be read in such a way as to
suggest that they MUST.

This commit simplifies the language and removes this possible
mis-reading.

-----
[View rendered criteria/open-licenses.md](https://github.com/publiccodenet/standard/blob/simpler-code-doc-license/criteria/open-licenses.md)

### DIFF
--- a/criteria/open-licenses.md
+++ b/criteria/open-licenses.md
@@ -6,7 +6,7 @@ order: 12
 
 ## Requirements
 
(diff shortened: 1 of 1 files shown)


clausmullie:
Great edit!

PR #305, opened 2020-02-27
Code coverage devs do

ericherman:
plugability for corner-case testing, and tuning code coverage tooling

### DIFF
--- a/criteria/continuous-integration.md
+++ b/criteria/continuous-integration.md
@@ -56,6 +56,7 @@ order: 11
 ## Developers and designers: what you need to do
 
 * Help management and policy makers test their contributions, by for example testing their contributions for broken links or style.
+* Structure code which is written to handle conditions which are difficult to create in a test environment (such as out of storage space or memory allocation failure) in such a way that the conditions can be simulated during testing.
 * Deploy often.
 
 ## Further reading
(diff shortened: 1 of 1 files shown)


ericherman:
>     1. The first bullet point is a very long nested sentence with double "which" and "such". Can it perhaps be split up into two sentences to make it easier to grasp?

What about this wording?

* Structure code which is written to handle conditions which are difficult to create in a test environment in such a way that the conditions can be simulated during testing. Forms of resource exhaustion such as running out of storage space or memory allocation failure are typical examples of difficult to create conditions.

Ainali:
Yes, that would be a lot easier to read!

ericherman:
or maybe:

* Structure code written to handle conditions which are difficult to create in a test environment in such a way that the conditions can be simulated during testing. Forms of resource exhaustion such as running out of storage space and memory allocation failure are typical examples of difficult to create conditions.

Ainali:
> or maybe:
> 
> * Structure code written to handle conditions which are difficult to create in a test environment in such a way that the conditions can be simulated during testing. Forms of resource exhaustion such as running out of storage space and memory allocation failure are typical examples of difficult to create conditions.

Even better!

PR #312, opened 2020-04-02
Issue #38, "civic code" -> "civil code"

ericherman:
https://github.com/publiccodenet/standard/issues/38

Use the more appropriate term "civil code" rather than "civic code"
in the glossary.

-----
[View rendered glossary.md](https://github.com/publiccodenet/standard/blob/eh-issue-38-civic-civil/glossary.md)


ericherman:
https://en.wikipedia.org/wiki/Civil_code

Issue #307, opened 2020-03-04
New requirement about intellectual property


Ainali:
During our brief governance meeting today @bvhme floated the idea of making it a requirement that intellectual property in public code projects should be retained by those making contributions. The argument for this is that many civil servants are prohibited by local national law to sign it away. I really like this idea and it would probably fit well in [Publish under an open license](https://github.com/publiccodenet/standard/blob/develop/criteria/open-licenses.md).

Ainali:
Is this a MUST or a SHOULD? I am leaning towards the former, but perhaps that is too strict?

bvhme:
I would say that as we're helping organizations to enable them to work with public organizations in different contexts: that disallowing a whole swath of them to contribute by creating an intellectual property structure that makes it illegal for them to contribute, seems like a bad idea.

So I think I'd opt for a `MUST` here.

ericherman:
See also CLA issue: https://github.com/publiccodenet/standard/issues/211
And related PR: https://github.com/publiccodenet/standard/pull/255

Ainali:
Solved by #313

PR #313, opened 2020-04-03
New requirement: retaining intellectual property

Ainali:
Solves #307 

-----
[View rendered criteria/open-licenses.md](https://github.com/publiccodenet/standard/blob/intellectual-property/criteria/open-licenses.md)

### DIFF
--- a/criteria/open-licenses.md
+++ b/criteria/open-licenses.md
@@ -9,7 +9,7 @@ order: 12
 * All code and documentation MUST be licensed such that it may be freely reusable, changeable and redistributable.
 * Software source code MUST be licensed under an [OSI-approved open source license](https://opensource.org/licenses/category).
 * All code MUST be published with a license file.
-* A contributor MUST be able to retain the intellectual property of their contributions to the codebase.
+* Contributors MUST NOT be required to transfer copyright of their contributions to the codebase.
 * All source code files in the codebase SHOULD include a copyright notice and a license header.
 * Codebases MAY have multiple licenses for different types of code and documentation.
 * Documentation MAY be published under [Creative Commons licenses](https://creativecommons.org/licenses/) that are NOT 'no derivatives' or 'non-commercial'.
--- a/criteria/open-licenses.md
+++ b/criteria/open-licenses.md
@@ -9,6 +9,7 @@ order: 12
 * All code and documentation MUST be licensed such that it may be freely reusable, changeable and redistributable.
 * Software source code MUST be licensed under an [OSI-approved open source license](https://opensource.org/licenses/category).
 * All code MUST be published with a license file.
+* A contributor MUST be able to retain the intellectual property of their contributions to the codebase.
 * All source code files in the codebase SHOULD include a copyright notice and a license header.
 * Codebases MAY have multiple licenses for different types of code and documentation.
 * Documentation MAY be published under [Creative Commons licenses](https://creativecommons.org/licenses/) that are NOT 'no derivatives' or 'non-commercial'.


bvhme:
I like the `be able to` part of this, because it clearly illustrates that it needs to be possible without forcing others to. There are different ways of solving this even with CLAs.

Ainali:
I thought intellectual property was the English term used to collectively address rights like copyright, trademarks, design protection and patents. At least we have a word for that in Swedish. Rights in general seems a bit too vague, whereas copyright is too narrow.

bvhme:
Is copyright too narrow? Come to think of it this specific point should maybe just be about copyright.

Is there a non-copyright example we can come to?

Ainali:
I was mostly thinking about patents. Let's say someone from for example Tesla (that have patents, but also releases them under a free license) wants to reuse such code in a repository. Should they have to sign away the patent rights?

bvhme:
That makes sense. So copyright and patents.

Perhaps we could ask @marctjones for help, since he's added to this
conversation before
<https://github.com/publiccodenet/standard/issues/211#issuecomment-504170513>
.

ericherman:
While the term "intellectual property" shows up in US law in a few places, like the DCMA I believe, typically what people mean by the vague phase is some combination of copyright, patent, trademark, and possibly other areas ... but what exactly is encompassed is often not clear. Part of the lack of clarity is that international treaties and conventions cover different elements of these things. Copyright is very uniform between countries, but patent, trademark, trade secret, are less so. I believe it is rare that something holds for two elements of "intellectual property" (for instance trade secret and copyright) without specifying a country.

I can't do the topic justice, but others have written about it:

Cory Doctorow: 
* "Sole and Despotic Dominion" https://locusmag.com/2016/11/cory-doctorow-sole-and-despotic-dominion/
* "Intellectual property" is a silly euphemism, https://www.theguardian.com/technology/2008/feb/21/intellectual.property

EFF:
* "Intellectual Property: The Term", https://www.eff.org/issues/intellectual-property/the-term

GNU:
* 'Did You Say “Intellectual Property”? It's a Seductive Mirage', https://www.gnu.org/philosophy/not-ipr.html
* "Don't Let ‘Intellectual Property’ Twist Your Ethos", https://www.gnu.org/philosophy/no-ip-ethos.html

 Rickard Falkvinge:
* "Language Matters: Framing The Copyright Monopoly So We Can Keep Our Liberties" https://torrentfreak.com/language-matters-framing-the-copyright-monopoly-so-we-can-keep-our-liberties-130714/

Ainali:
Alright with the term being so loaded in American English, perhaps we need to go for the wider term "rights" instead then.

bvhme:
I think that seems all right, however I'm a bit afraid it is harder to understand and might be construed to be about other things such as liabilities. Perhaps enumerating copyright and patents might be a better way.

ericherman:
I don't think "retain the rights to their contributions" would be construed to be about liabilities. But maybe? Since the licenses are copyright licenses, I'd suggest simply stating copyright, and skipping patent for now. I guess "rights covered by the license" might work, but that would likely make for an awkward and probably less clear sentence.

Ainali:
I don't think there ever has been a "right to be liable", it would take a pretty utilitarian society to phrase it like that, hence I am inclined to go for "retain the rights to their contributions".

bvhme:
Except for the complying with the License, which might be seen as not retaining your rights, or at least also extending them to others irrevokably. Maybe? I'm lost in this a bit.

ericherman:
Is it better to reformulate this as something like, "contributors MUST NOT be required to transfer copyright of their contributions to the codebase" ... would that sort of approach achieve the goal more directly?

Ainali:
> Is it better to reformulate this as something like, "contributors MUST NOT be required to transfer copyright of their contributions to the codebase" ... would that sort of approach achieve the goal more directly?

Yes. And it will cleanly solve #307 in how it was originally stated. My deviations with patents was perhaps a kind of feature creep.

PR #321, opened 2020-05-08
Add new terms and footnote section to glossary.md

felixfaassen:
I've added 3 terms which are used in the standard but where not on the glossary:

Version Control / Repository / Continuous Integration. Please feel free to rewrite as in software engineering the definition of these terms is much debated (like continuous integration for example)

I also added a footnote section to be able to reference to the original source material of the definition. Git Markdown does not support footnotes so I used the HTML anchors however I can imagine this way of handling footnotes is far from ideal (not "pure" markdown, and numbering issues) I will create a separate issue for handling footnotes to refer to original source material when re-using text or definitions from other material.

-----
[View rendered glossary.md](https://github.com/publiccodenet/standard/blob/felixfaassen-glossary-added-terms/glossary.md)

### DIFF
--- a/glossary.md
+++ b/glossary.md
@@ -12,7 +12,7 @@ This can be – for example – a document or a version-control repository.
 
 ## Continous Integration
 
-Continous Integration: In software engineering, continuous integration (CI) is the practice of merging all developers' working copies to a shared mainline of a codebase several times a day.<sup>[1](#footnote1)</sup>
+Continous Integration: In software engineering, continuous integration (CI) is the practice of merging all developers' working copies to a development branch of a codebase as frequently as reasonable. 
 
 ## General public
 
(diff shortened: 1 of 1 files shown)


ericherman:
Two nit-picks which should not prevent this change from coming in:

1) is there a less jargony word than "mainline"? Like "development branch"?

2) is there a better way to express that changes are merged as frequently as reasonable than to say "several times a day" ?  If changes are only at a rate of once per two days, it could still have/use CI.

bvhme:
I would like to have a discussion about the use of footnotes and this use of HTML in the MarkDown as it is introduced by this pull request.

Elsewhere we use inline links, blockquotes, text in brackets, and the 'further reading' section instead of footnotes.

We've decided on these as:

* They were easier to maintain as the content is logically bundled
* A smaller subset of markdown (without HTML elements) means it is easier for others to learn to contribute
* MarkDown doesn't really do footnotes well, and HTML also has no clear semantic way of creating and thus styling footnotes making it hard to transform content with footnotes with just CSS.

I personally fall on the side that footnotes don't make the content any clearer and I see it as a design solution, not a structural solution. Then again it's these kinds of edge cases that make CSS for print hell so I'm definitely biased as well :)

I'm curious as to your takes.

felixfaassen:
I removed the footnotes, added source inline in text and added the feedback from @ericherman

bvhme:
@ericherman I'm curious as to your (renewed) approval.

PR #328, opened 2020-05-12
Split requirement into brief example & test

ericherman:
I noticed that "SHOULD not" would be better as "SHOULD NOT" but
then realized that the requirement could be made sharper, and
some of the extra verbosity moved to the "How to test" section

-----
[View rendered criteria/reusable-and-portable-codebases.md](https://github.com/publiccodenet/standard/blob/eh-requirement-examples-as-test/criteria/reusable-and-portable-codebases.md)

### DIFF
--- a/criteria/reusable-and-portable-codebases.md
+++ b/criteria/reusable-and-portable-codebases.md
@@ -12,7 +12,7 @@ order: 3
 * The roadmap SHOULD be influenced by the needs of multiple parties.
 * Code SHOULD be general purpose and SHOULD be configurable.
 * Codebases SHOULD include a [publiccode.yml](https://github.com/italia/publiccode.yml) metadata description so that they're easily discoverable.
-* Code and its documentation SHOULD NOT contain situation-specific information such as hostnames.
+* Code and its documentation SHOULD NOT contain situation-specific information.
 
 ## Why this is important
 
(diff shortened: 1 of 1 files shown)


Ainali:
I like the split. Should we even phrase it like this: `Code and its documentation SHOULD NOT contain situation-specific information.` and move host names to the other enumeration as well to make it really clear cut? If not, I suggest to repeat host names in that enumeration to make it clear that they are connected.

ericherman:
https://standard.publiccode.net/criteria/code-in-the-open.html says:
> Contributors MUST NOT upload sensitive information regarding users, their organization or third parties to the repository. Examples of sensitive information include configurations, usernames and passwords, public keys and other real credentials used in the production system.

Here (in https://standard.publiccode.net/criteria/reusable-and-portable-codebases.html ) repeating users and passwords seems redundant, but an example -- such as hostname -- may be useful in conveying the idea that it's not only about PII and secrets, but about assumptions that make it harder for "reusers" to deploy/test

criteria/reusable-and-portable-codebases.md
```diff
@@ -12,7 +12,7 @@ order: 3
 * The roadmap SHOULD be influenced by the needs of multiple parties.
 * Code SHOULD be general purpose and SHOULD be configurable.
 * Codebases SHOULD include a [publiccode.yml](https://github.com/italia/publiccode.yml) metadata description so that they're easily discoverable.
-* Code and its documentation SHOULD not contain situation-specific information. For example, personal and organizational data as well as tokens and passwords used in the production system should never be included.
+* Code and its documentation SHOULD NOT contain situation-specific information such as hostnames.
```
bvhme, line 15: A hostname does seem like an odly specific example though to be the lone example and I quite some developers that would not really know what a hostname is, let alone managers and designers, although I would like them to make sure of this as well.

I'd say perhaps we can leave the example off until we find a more examplary example or add some more.

```suggestion
* Code and its documentation SHOULD NOT contain situation-specific information.
```

bvhme:
@ericherman, will you merge this?

Issue #269, opened 2019-09-10
Consider rewording license choice for documentation


Ainali:
The requirement in [Publish with an open license](https://standard.publiccode.net/criteria/open-licenses.html): "Documentation MAY be published under Creative Commons licenses that are NOT ‘no derivatives’ or ‘non-commercial" has a tricky wording and also puts the bullet point far away from the super bullet point that applies to both code and documentation. Something in the line with "Documentation MUST NOT be published under Creative Commons licenses with the permissions ‘no derivatives’ or ‘non-commercial" might be better.

bvhme:
I think that makes sense. In the end the goal of this is to prevent people choosing licences such as 'cc-by' or 'cc-nc' that restrict forking and continuous development of the codebase.

Ainali:
Just to clarify, you meant 'cc-nd' and 'cc-nc', right?

bvhme:
Definetely 'cc-nd' :)

And I'm wondering, wouldn't CC-BY also break forks, I thought it was incompatible with Open Source licenses, partly due to the requirement to indicate what you've changed. Although that seems pretty much inline with the version control requirement.

Also what about the alledged GPL incompatibility of CC-BY-SA? I don't seem to be able to wrap my head around this.

bvhme:
But that is not this core issue.

Ainali:
As far as I have understood it, the reason that CC BY and CC BY-SA is not used for software is that those were not designed for code (i.e. have nothing about patents etc.). They do qualify for the [open definition](https://opendefinition.org/licenses/). Note that OSI's site is under CC BY :) . 

CC BY-SA is compatible (in one direction) with GPL ([source](https://creativecommons.org/2015/10/08/cc-by-sa-4-0-now-one-way-compatible-with-gplv3/)).

Ainali:
When thinking of it more, I believe this is redundant with:

> All code and documentation MUST be published under a license that allows it to be freely reusable, changeable and redistributable

bvhme:
Enough so we can close this issue?

Ainali:
If you prefer me creating a new issue for the redundancy, then sure.

bvhme:
Sorry, I thought you thought the issue was redundant. And yes, I think the redundancy is true, this can be removed, however, I feel that this is still valuable guidance, so is there another place we could stash this?

On the original wording: I think it is important that we make clear we do think that creative commons licences are good, however, some licenses are not ok.

[...]

PR #334, opened 2020-05-18
Clarifying licenses for documentation

Ainali:
This moves the requirement to what management need to do.
Solves #269

-----
[View rendered criteria/open-licenses.md](https://github.com/publiccodenet/standard/blob/clarify-license-choice/criteria/open-licenses.md)


bvhme:
I like this!

However can we have the discussion about whether management is the group that picks the licenses? Are they in real life? Should they? Should we tell them to so that developers and designers dont?

PR #346, opened 2020-07-02
In use by mulitiple parties: MUST -> SHOULD

ericherman:
There can be some situations where public code is not
in use by multiple parties, thus relax from MUST to
SHOULD for this requirement.

-----
[View rendered criteria/reusable-and-portable-codebases.md](https://github.com/publiccodenet/standard/blob/eh-multi-party-must-to-should/criteria/reusable-and-portable-codebases.md)

### DIFF
--- a/criteria/reusable-and-portable-codebases.md
+++ b/criteria/reusable-and-portable-codebases.md
@@ -8,7 +8,7 @@ order: 3
 
 * The codebase MUST be developed to be reusable in different contexts.
 * The codebase MUST be independent from any secret, undisclosed, proprietary or non-open licensed code or services for execution and understanding.
-* The codebase MUST be in use by multiple parties.
+* The codebase SHOULD be in use by multiple parties.
 * The roadmap SHOULD be influenced by the needs of multiple parties.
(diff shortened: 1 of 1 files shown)


Ainali:
I am not sold yet. It feels artificial.

Ainali:
> I am not sold yet. It feels artificial.

And by artificial I mean, the change makes more codebases pass the Standard without adding value to ecosystem. To extrapolate in this direction, one could argue that we should remove the requirement completely.

ericherman:
There are instances where we would believe Public Code should be considered to meet the Standard even though it is not yet in re-use.

One example is that if as part of procurement for a new codebase, the developed code must meet the criteria of the Standard, then this early bootstrapping period would not yet have re-use in multiple contexts.

Some policy applies to a very limited set of possible deployments. Hypothetically, something developed for the European Parliament might be potentially reusable by other groups of nations with similar policies and goals, yet we may not yet see the reuse within this small set of potential contexts.

Ainali:
That is a good argument. Also, it would be useful for public bodies to be able to write into any contract that what's delivered should comply with the Standard (and that contractor should then not need to go get other adopters of the codebase).

PR #352, opened 2020-09-24
Add 'Make contributing easy' to the index.

bvhme:
It was missing, which is a shame because it's great.

We forgot to update the index when splitting the criteria.

-----
[View rendered index.md](https://github.com/publiccodenet/standard/blob/fix-missing-index/index.md)


ericherman (approved):
D'oh. What an oversight! Yes. This should have been part of the earlier PR.

PR #366, opened 2020-10-26
Add Markdown linting to script/test.sh

ericherman:
The develop branch currently contains many linting errors,
we should examine and resolve those one-by-one and remove them
from the exceptions list.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,4 @@
 source "https://rubygems.org"
 gem 'github-pages', group: :jekyll_plugins
 gem 'html-proofer'
+gem 'mdl'
--- a/script/test.sh
+++ b/script/test.sh
@@ -1,6 +1,22 @@
 #!/usr/bin/env bash
 set -e # halt script on error
 
+# Lint markdown using the Markdownlint gem with the default ruleset except for:
+# MD007 Unordered list indentation: we allow sub-lists to also have bullets
+# MD013 Line length: we allow long lines
+# MD029 Ordered list item prefix: we allow lists to be sequentially numbered
+#
+# Additionally, we have these violations which should be resolved:
+# MD004 Unordered list style
+# MD009 Trailing spaces
+# MD012 Multiple consecutive blank lines
+# MD022 Headers should be surrounded by blank lines
+# MD026 Trailing punctuation in header
+# MD032 Lists should be surrounded by blank lines
+# MD034 Bare URL used
+#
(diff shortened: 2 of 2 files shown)


ericherman:
This should not target the current 0.2.0 release which is in-progress, rather it should come after.

bvhme (approved):

bvhme:
I propose merging this and creating an issue to track resolving the missing tests so they can be fixed on a per-test basis, or as useful.

Issue proposal:

`
Re-enable the following MarkDownLint rules:

- [ ] Additionally, we have these violations which should be resolved:
- [ ] MD004 Unordered list style
- [ ] MD009 Trailing spaces
- [ ] MD012 Multiple consecutive blank lines
- [ ] MD022 Headers should be surrounded by blank lines
- [ ] MD026 Trailing punctuation in header
- [ ] MD032 Lists should be surrounded by blank lines
- [ ] MD034 Bare URL used
`

Ainali:
I am not sure we want to resolve `MD026 Trailing punctuation in header`, that should perhaps go to the permanent list of exceptions.

bvhme:
I definitely think trailing punctuation in headers is quite an important one as it prevents the use of headers as content. An often-made mistake is that people put questions in headers, confusing structure and flow. This rule guards against that.

Ainali:
> the use of headers as content

I am not sure I understand what this means.

bvhme:
Headers serve as a way to structure and navigate content. A header tells you what content is following so that you can find it later and can understand where you are in the structure. This is a lot easier and requires a lot less cognitive overhead if those headers are descriptive instead of elusive.

The body text contains the content and it has 'flow'. Headers provide structure there but they should not interrupt flow. Many people even skip over headings in long articles.

One should be able to read the content aloud without having to read the headers, and the headers should not be the only place where any crucial content is hidden. In addition, one should be able to read the headers and understand where to jump in when they're looking for something.

This is clearly explanatory, tells you what's underneath, and can also be made up if not read itself:

> Cats don't like to be near dogs

While reading the following leaves you wondering what's underneath, if it has the information you're looking for and likely is also an integral part of the content:

> Do cats not like to be near dogs?

I make this error all the time. For me, that largely stems from my writing process. I start with an outline, fill that out. This means that often my outline items turn into the headers. This makes sense in my writing process but not for readers. So I often need to go over my headers and make sure they are actually headers.

bvhme:
In addition to this, any punctuation in a header terminates a statement and calls for a rest. This is not a role for headers and can impede the readers' natural rythm.

ericherman:
Added issue https://github.com/publiccodenet/standard/issues/373 to cover the exceptions required to get this first step merged.

PR #372, opened 2020-10-27
Add the title to the title page

bvhme:
As it stands we need to generate a separate cover for the PDF.
This change adds the title to the title page.
This means the generated PDF can be used without requiring manual generation.

image

### DIFF
--- a/print.html
+++ b/print.html
@@ -105,6 +105,13 @@
       clear: both;
     }
 
+    h1.title {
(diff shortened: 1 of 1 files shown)


ericherman (approved):
This change works for making the printed version's inside first page more clear as well as having the contents work well without being wrapped in a front-and-back-cover pdf.

PR #380, opened 2020-10-28
Ignore "429 Too Many Requests" link-check failures

ericherman:
Our CI often fails with HTTP status code 429, Too Many Requests.
While a 429 might hide a 404 this has not been a problem for us yet.

We can separately check all links and alert on failures which were
hidden by ignoring 429 errors if we wish.

See also: https://github.com/publiccodenet/standard/issues/344


ericherman:
If we merge this (as I propose we should), I suggest we *also* create an issue to monitor possible link errors hidden by ignoring 429 errors in this test.

bvhme (approved):

PR #416, opened 2020-11-26
Specify intent of choice of file formats

### LINKED ISSUE #316: Version control or compare changes?
The criteria [Maintain version control](https://standard.publiccode.net/criteria/version-control-and-history.html) has the requirement:

> You SHOULD prefer file formats that can easily be version controlled.

Technically, modern version control software can handle most files with ease. But what we do want to say with this is more precise, we want to be able to compare changes easily, and not only on a technical level, but on a human readable level. Can we make this more clear in the requirement or is this nitpicky and actually obvious?

Ainali:
Fixes #316

-----
[View rendered criteria/version-control-and-history.md](https://github.com/publiccodenet/standard/blob/ja-meaningful-diffs/criteria/version-control-and-history.md)

### DIFF
--- a/criteria/version-control-and-history.md
+++ b/criteria/version-control-and-history.md
@@ -12,7 +12,7 @@ order: 6
 * Every commit message MUST link to discussions and issues wherever possible.
 * You SHOULD group relevant changes in commits.
 * You SHOULD mark different versions of the codebase, for example using revision tags or textual labels.
-* You SHOULD prefer file formats that can easily be version controlled.
+* You SHOULD prefer file formats where the changes of the files can be easily viewed and understood in the version control system.
 
 ## Why this is important
 
--- a/criteria/version-control-and-history.md
+++ b/criteria/version-control-and-history.md
@@ -12,7 +12,7 @@ order: 6
 * Every commit message MUST link to discussions and issues wherever possible.
 * You SHOULD group relevant changes in commits.
 * You SHOULD mark different versions of the codebase, for example using revision tags or textual labels.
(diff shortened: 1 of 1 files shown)


ericherman (approved):
Is "of" the best word?

Would it be better as "changes *in* the files" or perhaps "changes *within* the files" ....

Regardless of preposition choice, the wording in this diff is an improvement.

Ainali:
I defer to native speakers here and are happy to change. @ElenaFdR any preference between 'in' or 'within'?

criteria/version-control-and-history.md
```diff
@@ -12,7 +12,7 @@ order: 6
 * Every commit message MUST link to discussions and issues wherever possible.
 * You SHOULD group relevant changes in commits.
 * You SHOULD mark different versions of the codebase, for example using revision tags or textual labels.
-* You SHOULD prefer file formats that can easily be version controlled.
+* You SHOULD prefer file formats where the changes of the files can be easily viewed and understood in the version control system.
```
ElenaFdR, line 15: 2 options, with slightly different nuance. The first probably fulfills #316 slightly better, but is a clunky sentence.

```
* You SHOULD prefer file formats where the changes within the files can be easily viewed and understood in the version control system.
```
or:

```
* You SHOULD prefer file formats where changes can be easily viewed and understood in the version control system.
```

ericherman: Yeah, I see what you mean. For me, I think the "clunkiness" is probably worth it for the explicitness of fulfilling #316.

ericherman (approved):

PR #420, opened 2020-12-03
Change "feedback on all decisions" to "feedback on all decisions to not accept a contribution"

### LINKED ISSUE #349: "Reviewers MUST provide feedback on all decisions" lacks clarity of intent
Requirement text is "Reviewers MUST provide feedback on all decisions made and the implementation in the review."

If this is intended to deny "Looks Good To Me" as a valid review, then this text is correct, however in that case I suggest a SHOULD rather than a MUST.

If the intent is to ensure feedback for any reasons that a reviewer may *reject* a review, then I am on-board with the MUST level of requirement and we should clarify the text to reflect that nuance.

ElenaFdR:
Closes #349.

-----
[View rendered criteria/require-review.md](https://github.com/publiccodenet/standard/blob/decision-to-not-accept/criteria/require-review.md)

### DIFF
--- a/criteria/require-review.md
+++ b/criteria/require-review.md
@@ -8,7 +8,7 @@ order: 7
 
 * All contributions that are accepted or committed to release versions of the codebase MUST be reviewed by another contributor.
 * Reviews MUST include source, policy, tests and documentation.
-* Reviewers MUST provide feedback on all decisions made and the implementation in the review.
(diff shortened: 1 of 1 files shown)


raboof (approved):
makes sense to me :+1:

ericherman (approved):

Ainali (approved):

Issue #419, opened 2020-11-30
Standard lacks guidance for "rolling release" environments


ericherman:
Versioning as described in criteria/document-maturity.md and criteria/version-control-and-history.md is clear for packaged software and APIs, but a little less clear for codebases which have a rolling-release process especially where the updates tend to involve regular content changes or evolving layout with A/B testing.

Even if the criteria requirements are sufficient, the guidance for these environments is thin.

Ainali:
[This is an example](https://poit.bolagsverket.se/poit/PublikStartIn.do) of a government site which _might_ be best suited to rolling releases since it is for public service announcements.

ElenaFdR:
@Ainali I'd expect rolling content changes to be handled by a CMS, not in the code.

raboof:
> @Ainali I'd expect rolling content changes to be handled by a CMS, not in the code.

While a CMS is a fine tool for rolling content changes, using a version control system (and a 'continuous delivery' pipeline to let changes flow to production, for example with a static site generator like jekyll or hugo) is also a great approach.

Also for projects (less so for libraries) I think this is a valid approach. I still think, even if artifacts are produced for each merge to master, those artifacts should be versioned somehow. Perhaps we should just clarify that it's OK if versions are derived from the version control system metadata (like git-describe) - something like #421?

ericherman:
Resolved: https://github.com/publiccodenet/standard/pull/421

PR #421, opened 2020-12-04
Clarify versions may be derived from VCS metadata

raboof:
In continuous deployment/rolling release scenario's, especially
for 'leaf' projects (less so for libraries) it is still important
to have a clear identifier for versions. However, rather than
manually tagging each version, the version would be derived from
the version control system metadata. It might be good to clarify
that explicitly.

References #419

-----
[View rendered criteria/document-maturity.md](https://github.com/raboof/standard/blob/allow-rolling-updates/criteria/document-maturity.md)

### DIFF
--- a/criteria/document-maturity.md
+++ b/criteria/document-maturity.md
@@ -41,6 +41,7 @@ Clearly signalling a codebase's maturity helps others decide whether to reuse, i
 
 * Add a prominent header to every interface that indicates the maturity level of the code.
 * Version all releases.
+* Especially in 'rolling release' scenario's, the version may be automatically derived from the version control system metadata (e.g. using [git describe](https://git-scm.com/docs/git-describe)).
 
 ## Further reading
 


ericherman:
I think "manual" is unlikely in a continuous delivery setup. When I worked in a rapid delivery environment, the automated rollout process created the tag. Tags to the specific commit included the date-time of tag creation. I expect that is common for rapid release contexts, but my experience is limited.

Let's think about what guidance could be added (perhaps in "Developers and designers: what you need to do") ... please jot-down a suggestion if one comes to mind.

raboof:
> I think "manual" is unlikely in a continuous delivery setup.

I completely agree

> the automated rollout process created the tag

Indeed. Another approach I see a lot is where 'incremental' deployments are not explicitly tagged in VCS at all, and get version numbers that are derived from VCS (`git describe`-style) - that way the deployment machinery doesn't need write access to VCS which is neat.

I'll amend the PR to clarify.

> Let's think about what guidance could be added (perhaps in "Developers and designers: what you need to do") ... please jot-down a suggestion if one comes to mind.

I'm not sure we should be opinionated here, as long as each deployment can be traced back to a specific version of the code I'm happy :)

ericherman:
Apologies for the long delay. Each time I circle back to this I get a bit stuck thinking about it.

My thinking is that this is not a "MAY" requirement of the criterion. I agree that "a codebase MUST be versioned" would be achieved by a process which automatically tags it in the version control system metadata, however this is an implementation option rather than a new line-item requirement.

I would happily support this idea it were text added to the "Developers and designers: what you need to do" or "How to test" sections, as it could certainly address Issue #419 from either of those locations.

@felixfaassen @AlbaRoza and @Ainali ... any thoughts you wish to chime-in with?

Ainali:
I agree. It is good advice specifically when handling rolling releases as in #419, but as a MAY this suggests that we think it is a good idea for all codebases.

ericherman:
Good observation @Ainali, the other "MAY" requirements are more like "Suggested" than "Permitted" for the most part.

raboof:
> as a MAY this suggests that we think it is a good idea for all codebases

I don't think MAY should be read like that - we're deriving the terminology from the RFC and while https://tools.ietf.org/html/rfc2119#section-5 doesn't say so explicitly, it is often used in the 'permitted' sense rather than the 'suggested' sense (for which there is also SHOULD).

> I would happily support this idea it were text added to the "Developers and designers: what you need to do" or "How to test" sections

This might still be a good idea - I think 'what you need to do' fits best, moved it there

Ainali:
> This might still be a good idea - I think 'what you need to do' fits best, moved it there

Nice, I like it!

Just an ending period in the sentence, and we are good to go!

Ainali (approved):

raboof:
> Just an ending period in the sentence, and we are good to go!

:+1: done

ericherman (approved):

PR #426, opened 2021-01-11
Replace unreliable bpmn.org link

ericherman:
As http://www.bpmn.org/ has not been reliable, this converts the link to
https://en.wikipedia.org/wiki/Business_Process_Model_and_Notation

-----
[View rendered criteria/bundle-policy-and-code.md](https://github.com/publiccodenet/standard/blob/eh-fix-bpmn-link/criteria/bundle-policy-and-code.md)


Ainali (approved):
The link works, and the change is an improvement. As a sidenote, the Wikipedia page has the "This article has multiple issues." tag, so we should really try to dig for some good sources and improve the article.

bvhme (approved):
It's always sad that sites like these drop off the internet. However this is good, and might even be an improvement :)

PR #427, opened 2021-01-11
Ghactions test 202101

ericherman:
Actions seem to be working fine!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,10 +4,9 @@ on: push
 defaults:
   run:
     shell: bash
-    # working-directory: scripts
 
 jobs:
-  job1:
+  cibuild:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,21 @@
+name: Test
+on:
+  push:
+    branches:
+      - ghactions-test-202101
+
+defaults:
+  run:
+    shell: bash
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,18 @@
+name: Test
+on: push
+
+defaults:
+  run:
+    shell: bash
+    # working-directory: scripts
+
+jobs:
+  job1:
```
bvhme, line 10: Can we give this job a more telling name like 'cibuild' or something :)

.github/workflows/test.yml
```diff
@@ -0,0 +1,18 @@
+name: Test
+on: push
+
+defaults:
+  run:
+    shell: bash
+    # working-directory: scripts
```
bvhme, line 7: This can probably be removed

bvhme (approved):

.github/workflows/test.yml
```diff
@@ -0,0 +1,17 @@
+name: Test
+on: push
+
```
felixfaassen, line 3: @ericherman we could consider also triggering the test on a pull request (on a certain branch):  
```
on: push
      pull_request:
          branches:
             - master
```

felixfaassen (commented):
We could consider adding a trigger condition for the test on pull_request (see comment on the test.yml. But good also do it later. Not a blocker ;-)

ericherman:
I'll merge this as a start and then we can add  triggering the test on a pull request separately. Maybe we would want to do this on any PR to master or develop?

PR #437, opened 2021-02-03
Fix reference in glossary

### LINKED ISSUE #403: Source retrieved 6 years ago?
In the [glossary](https://github.com/publiccodenet/standard/blob/develop/glossary.md), we state that the source for Version control was retrieved 6 years ago. Did we actually retrieve it back then and kept in memory until now, or is this a [copy and paste error](https://en.wikipedia.org/wiki/Continuous_integration#cite_note-martinfowler-1)?

Ainali:
Restyled it and made the intro start with the gloss.

Fixes #403

-----
[View rendered glossary.md](https://github.com/publiccodenet/standard/blob/ja-fix-glossary-reference/glossary.md)


glossary.md
```diff
@@ -61,4 +61,4 @@ In revision (or version) control systems, a repository is a data structure which
 
 ## Version control
 
-A component of software configuration management, version control, also known as revision control or source control, is the management of changes to documents, computer programs, large web sites, and other collections of information. Changes are usually identified by a number or letter code, termed the "revision number", "revision level", or simply "revision". For example, an initial set of files is "revision 1". When the first change is made, the resulting set is "revision 2", and so on. Each revision is associated with a timestamp and the person making the change. Revisions can be compared, restored, and with some types of files, merged. (source: Fowler, Martin (1 May 2006). "Continuous Integration". martinfowler.com. Retrieved 9 January 2014.)
```
raboof, line 64: can we use this text without attribution copyright-wise?

Ainali: Well, the text is really from Wikipedia, so I don't think it was verbatim copied from the source anyway. If anything, we should acknowledge it.

ericherman (approved):

glossary.md
raboof: Wikipedia also requires attribution, though, right?

Ainali: Yes, resolving this by rewriting it completely.

ericherman (approved):
It's fine as-is. It's probably better to state that revision control makes things "easier" rather than "easy", but that is a nit-pick and should not block this.

Ainali:
> It's fine as-is. It's probably better to state that revision control makes things "easier" rather than "easy", but that is a nit-pick and should not block this.

That's an excellent point! I'll change that now.

ericherman (approved):

PR #439, opened 2021-02-07
Use standard 'advertise' in favor of 'advertize'

### LINKED ISSUE #400: Standardize 'advertise' vs 'advertize'
We currently use both in the Standard, which isn't ideal but doesn't impede understanding.

Huy-Ngo:
This pull request means to resolve #400.

The current situation is that both "advertize" and "advertise" are used. According to the links mentioned in the issue ([Grammarist](https://grammarist.com/spelling/advertise-vs-advertize/) and [StackExchange](https://english.stackexchange.com/questions/341707/should-advertised-be-spelled-with-a-z-in-american-english)), the former is not standard and we should use "advertise" instead.

-----
[View rendered criteria/open-to-contributions.md](https://github.com/Huy-Ngo/standard/blob/develop/criteria/open-to-contributions.md)

### DIFF
--- a/criteria/open-to-contributions.md
+++ b/criteria/open-to-contributions.md
@@ -8,7 +8,7 @@ order: 4
 
 * The codebase MUST allow anyone to submit suggestions for changes to the codebase.
 * The codebase MUST include contribution guidelines explaining how contributors can get involved, for example in a `CONTRIBUTING` file.
-* The codebase SHOULD advertize the committed engagement of involved organizations in the development and maintenance.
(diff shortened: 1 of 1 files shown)


bvhme (approved):
Great contribution! I think this is ready to be merged.

I've checked there are no other mentions of `advertize`. Fixes #400.

@publiccodenet/codebase-stewards anyone up for merging this?

lscheske:
I know this has been merged, but we use American English spelling throughout, so we should actually use the version with a "z" rather than an "s".

PR #440, opened 2021-02-08
Eh check only new links

ericherman:
Here a rough starting point for checking new links.

TODO: At the time of this rough draft, it does not yet have a url-ignore feature, but I'd like to get some early feedback on the idea.

### DIFF
--- a/script/check-new-links.sh
+++ b/script/check-new-links.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+NOT_OK_COUNT=0
+SKIP_PATTERNS=(
+	'github\.com' # links may not exist yet
+	'twitter\.com' # twitter is broken
+)
+
+if [ "_${VERBOSE}_" == "__" ]; then VERBOSE=0; fi
+
+for URL in $(git diff develop |
+		grep '^+' |
+		grep -Eo '(http|https)://[^)">]+'); do
+	SKIP_URL=0
+	for PATTERN in "${SKIP_PATTERNS[@]}"; do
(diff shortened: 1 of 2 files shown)


Ainali:
I tried it and did a small change and got this result, which we perhaps should handle:

```
./script/check-new-links.sh 
https://help.github.com/articles/creating-an-issue/
HTTP/2 301 
FAIL: 1 new URLs were not OK
```

ericherman:
Perhaps. I'd like newly added links should not reference "moved permanently" old locations, thus I'm on the fence about this one.

Ainali:
That is a very good point and I think I agree with it in general, but doesn't that mean that some short URLs cannot be added?
I am not saying it is a blocker of any kind, just making sure we are making these decisions consciously.

ericherman:
@Ainali Your question raises a solid point ... perhaps "moved permanently" should maybe warn, but should not fail if the resource is there. Please try with this new addition.

Ainali (approved):
This is really cool and will be very helpful!

ericherman:
Okay, good. Thus the next step is to add a way to ignore urls. I'll sketch that out next.

ericherman:
> As a sidenote this example also shows why I (and we) should use new line more often, because the failing URL was not part of my change, but since the paragraph is long, it _is_ part of the diff.

Yes.
I think we can all be better about putting in newlines after every period.

ericherman:
Okay, URL ignore via regular expression added. @Ainali what do you think now?

Ainali (approved):
I tried a few different test cases and it works fine!

felixfaassen (approved):

PR #480, opened 2021-03-05
First page for generated PDF that can serve as the cover

bvhme:
Here is the new page 1 of the generated PDF file:

image

I haven't gotten to generating the print version yet, however this should remove the need to manually attach a cover file to new versions.

The separate cover file for the printer I'll make next, however I need to invent a way to do effective layout for it first with it's two pages in one style. It's in the works.

Also, there is no date on there, as we don't really have a system for that yet.

Should I add a published date to the `config.yml` that needs to get updated with the version?


ericherman:
Yes, date in `_config.yml` under `title` and `version` seems reasonable to me.

ericherman (approved):

ericherman:
Do you wish to merge this as-is or wait until add the date is added?

bvhme:
I might not be able to do any real work on this the next 14 days, so you could merge it as is and we can add the date later, wait (we're not in a rush right), or add it yourself.

ericherman:
let's merge and then add the date.

PR #481, opened 2021-03-05
PDF for the outer cover

bvhme:
Generates a PDF version of the outer cover of the Standard from Jekyll.

image


ericherman (approved):
Nice to see this, thank you.

Not a blocker: the `.gitignore` lacks a newline at the end of the file, this will make the diff "noisy" on the next addition to the file.

ericherman (approved):

PR #487, opened 2021-03-25
Extract markdown as separate GH Action step

ericherman:
This moves the "bundle exec mdl" into a stand-alone script.
The stand-alone script is now refrenced as a separate "run"
command in both the regular test and cron-nightly actions.
As .travis.yml still references cibuild.sh, it is called
form that script as well.
Developers may still find cibuild.sh handy as a "one stop"
for testing even after we remove .travis.yml from the repo.

### DIFF
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -25,4 +25,5 @@ jobs:
         with:
           ruby-version: '2.7'
       - run: bundle install
+      - run: ./script/test-markdown.sh
       - run: ./script/test-with-link-check.sh
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,4 +16,6 @@ jobs:
         with:
           ruby-version: '2.7'
       - run: bundle install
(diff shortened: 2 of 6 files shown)


felixfaassen (approved):
Hi @ericherman great addition. Small remark not for now maybe but another solution (which will give us more control about when and how to run certain steps) is to define it as a job  in stead of having multiple and separate yml files. This way we have more control and less redundancy of workflow files (eg when we need to bump up version numbers of ruby), have the same build environment and trigger jobs based on dependent other job results or have run different jobs based on merges on pulls. More info on that here: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobs

ericherman:
Conceptually, we have two sets of (distinct yet similar) tasks. If we can express that in a single workflow .yml file which defines both the jobs to run on each "push"/"pull request" event as well as the cron-like nightly "schedule" event, then that might allow us to reduce some duplication.

Let's experiment with that (maybe on a different repo) next week.

Issue #62, opened 2019-04-10
Bundle policy and code: invite policy makers to contribute their own work in their own formats


ElenaFdR:
'Bundle policy and code' encourages policy experts to use flowcharts (effectively) to chart the policy, but doesn't say anything about the formats for presenting policy that are more commonly used by policy people - probably something more like a long article or essay, potentially with graphs or diagrams.
Suggestions:
- Invite them to include their original policy papers (the work they did for their peers) so long as there's an English language translation
- Invite them to link to any explanations of the policy produced by the public administration for comms purposes (for example, for residents, or to sell the new policy politically)
- Make sure there's a cross reference from here to policy documentation in 'document your objectives'

ElenaFdR:
Additional thought: where either the code or the policy have to reference something that's specific to that policy environment (especially if in a non-English language), please include a brief English intro (eg Wikipedia link) to the concept.

ericherman:
@ElenaFdR when you say "invite" do you envision a MAY criteria or guidance text in the "Policy makers: what you need to do" section?

Ainali:
Let's plan a working session on this topic. We've recently gotten some feedback that also addresses this that we can incorporate.

Ainali:
Addressed by #504 and #505. If something is still missing, please create a new issue more narrow issue.

PR #504, opened 2021-06-14
Rephrase Document Objectives to highlight policy objectives

ericherman:
Rather than phrase the requirement in terms of "understandable by policy
makers", this phrasing connects the codebase goals to policy goals.

See also issue #62 

-----
[View rendered criteria/document-objectives.md](https://github.com/ericherman/standard-for-public-code/blob/policy-objectives-20210614/criteria/document-objectives.md)


Ainali (approved):
Nice! I think this better capture what we want to have documented, the clear line from the public policy to the code.

ericherman:
I suggest that whatever the wording, it must acknowledge that not all policy objectives are reflected in codebase, and similarly not all codebase objectives are going to be clear reflections of policy.

With that in mind, would wording similar to:

`Codebase documentation SHOULD clearly describe how policy objectives and codebase objectives are reflections of each other or connected.`

have lower "wedge" risk?

Ainali:
> The intent here, in my understanding, is to say that developers need to be able to describe what their code does to people who write code in a different language (policy makers) who are dependent on the code.

If this was the intent, perhaps we should keep this requirement (but making it clearer to make that intent clear)? 
What we tried to address was to make connection between the objectives of a policy (example: "The city should be a pleasant place for bikers and pedestrians") and a codebase objective (example: "The image algorithms should discover all cars parked on a bike lane or footpath"). The policy in itself have no objective of discovering cars, but how the codebase works is totally dependent on it.

bvhme:
I think you are right @Ainali, the challenge is indeed on how it is phrased in a way that the one is not excluding, since it is a bit grey-area all around.

In your example, the algorithm is also a form of policy, namely a computer run execution policy in service of a more abstract policy.

I suspect we need to create the space for the entire stack of abstractions of code to be documented instead of splitting them in the highest (laws/resolutions) and lowest level form (execution policy and software) without the intermediate levels.

felixfaassen (approved):

ericherman:
Would something like this be better?

```
The codebase SHOULD contain clear documentation of its objectives. Where applicable, codebase objectives should be expressed in terms of policy objectives.
 ```

**Comment by Ainali** [comment]
> Would something like this be better?
> 
> ```
> The codebase SHOULD contain clear documentation of its objectives. Where applicable, codebase objectives should be expressed in terms of policy objectives.
> ```

I don't know, this feels a bit mysterious to me and I am not really sure what it means. I guess what we want, but perhaps in a not so wordy phrasing is something like this:

```
The codebase SHOULD contain clear documentation of its objectives.
Especially, the documentation should explain how the codebase aims to solve the objectives of the public policies causing the creation of it.
```

ElenaFdR (approved):
Clear enough; I look forward to seeing how people act on it.

PR #513, opened 2021-08-16
Linkcheck: specify the minute

ericherman:
Specify a few minutes after the hour.


ElenaFdR (approved):
What advantage does this provide?

ericherman:
> What advantage does this provide?

First, while not specifying a minute is supported, it is more clear to specify a specific minute. Second, my hope is that some of the links which are less reliable will become more reliable if we specify a minute to fetch which is a few minutes after the top of the hour.

Issue #519, opened 2021-08-23
Add Digital Public Goods badge


Ainali:
We should add to the Standard that it is approved as Digital Public Goods. On the [printed version](https://github.com/publiccodenet/standard/blob/665e929f0c4bebea049915b2b096b44d52cc84a9/print.html) it would be fitting on the second page under the URL (line 284).

Where would it make sense in the online version?

Edit: 
- [x] Add badge on print.html
- [ ] Add badge in footer

ericherman:
Maybe in the footer with version, copyright, license, etc.

Ainali:
I believe the footer is entirely [footer-liquid](https://github.com/publiccodenet/jekyll-theme/blob/master/_includes/footer.liquid) in our Jekyll theme. How do we add something that should only show up in the footer on the Standard and not our other sites?

PR #520, opened 2021-08-23
Add DPG badge to print version

Ainali:
Works towards #519.


ericherman (approved):
I'm happy to get the print version in even before we decide where to add it to the on line version.

Issue #425, opened 2021-01-04
Is the introduction still fit for purpose?


ElenaFdR:
Does it still work for new readers? Is it still an accurate reflection of our aspirations for the Standard? And more importantly, does it reflect everything we've learned about how people (want to) use the Standard in the intervening 1.5+ years since it was written?

People have already raised separate issues about specific parts of the introduction:

- whether the Foundation's stewardship process needs such a detailed explanation [#396]
- making the introduction more relevant to the audiences addressed in the rest of the Standard [#280]
- adding a clearer values based entry point to the Standard [#318]
- ambiguous support for commercial licence models [#185]
- and most damningly, bad writing [#216]

One additional logistical reason to update it now is from the discussions on translating the Standard. We've discussed encouraging widespread translation of the introduction, but leaving the criteria in English (so that there's no ambiguity in the functional requirements). If widespread translation of the introduction is to have any of the impact we hope, then the intro needs to be the strongest, tightest, most compelling version of itself that it can be.

Ainali:
In the last community call we discussed this issue and here is [the blog post](https://blog.publiccode.net/community%20call/2021/01/25/notes-from-community-call-7-january-2021.html) that summarized the call.

TL;DR We will start by restructuring the introduction and moving around a few paragraphs.

ericherman:
Improving the introduction to match the more current landscape means changing the tone from "call-to-arms of a new world" to one of supporting policy goals of current digital transformations.

ericherman:
The Foundation's stewardship process is located at https://publiccode.net/standard-for-public-code/ and not duplicated in the standard.

By transforming the Introduction into a Foreword, this removes the dissonance in the tone from what makes sense in the text of a standard.

There is still room for transforming the reader's guide to an actual introduction relevant to reader's of the standard, but perhaps that should be its own issue.

A value-based foreword would certainly be an improvement, that should be a part of https://github.com/publiccodenet/standard/issues/216 I think.

ericherman:
I have a proposal for how I would like to hack on this, ideally co-working in one of our [community calls](https://community.standardforpubliccode.org/):

1. brain-storm elements we think a foreword should contain (any order)
2. review the Foundation foreword to see if there are any elements missed.
3. create a rough outline
4. transform outline to real text, aiming for a voice that better reflects the current community: more "sober" and "matter of fact" not a "call to arms" with "we believe" statements

Issue #218, opened 2019-06-02
introduction/certifying an entire codebase: unclear paragraph


Vesuvium:
```
If the audit process is added to an existing codebase, the new merge requests can be certified, but the existing code cannot be certified. By auditing every new merge request the codebase can move towards being completely certified.
```
I found the last sentence unclear because it does not say how many changes or how much time non-certified codebases need to be completely certified.

I would think that a non-certified codebase can become certified even if it's not entirely audited, for example after 6 months of certified pull requests or when 50% of the codebase has been certified, because I know that new pull requests are unlikely to cover 100% of a codebase in a timely manner.

That would be a bad practice, so I imagine at somepoint a full audit will be made, but the paragraph hints that this does not happen and does not state it clearly.

clausmullie:
While it is true that new pull requests are unlikely to cover all of the code (eg initial base functionality that works perfectly fine) but we could require a full peer-reviewed refactor (or at least check) of all existing code before certification.

This seems important to, for example, ensure consistent high quality documentation throughout the codebase. 

Do you feel this would be reasonable/feasible? If so, any suggestions for how to word this?

Vesuvium:
With a small codebase or with a well-written one, you can certainly do that. The problem I see is with large codebases with poor code quality and maintainability, where it's going to be too expensive or impossible to review everything.

> By auditing every new merge request the codebase can move towards being completely certified.

Then this sentence becomes "illusionary", since that's never going to happen. And it conflicts with ` the existing code cannot be certified`.

What could work is adding an exception to that, by saying that small codebases are going to be reviewed and certified. For example:

```
If the audit process is added to an existing codebase, the new merge requests can be
certified, but the existing code cannot be certified. 

Existing codebases can be reviewed and certified when they meet these criterias:

- A coverage higher than X%
- A line count less than Y
- A mantainability metric below Z
```

ericherman:
If this text content moves from the body of the Standard to the publiccode.net, the issue should be tranfered as well.

Ainali:
> I found the last sentence unclear because it does not say how many changes or how much time non-certified codebases need to be completely certified.

I think the key is in the first sentence in the same paragraph:

> For the codebase to be completely certified every meaningful line of code, and the commits behind the code, need to meet the Standard.

[...]

PR #521, opened 2021-08-23
Start refactoring the introduction

Ainali:
This pull request is intentionally bold to get our discussion more concrete. Even though suggestions of changes to pull requests are always welcome, in this one they are especially appreciated.

Since some texts may fit better elsewhere, I removed:
- Why public code?
- How public code works

I moved the section with the goals ahead of the section with certification process, since that is more of an overview, especially with the other text gone.

I also removed the Further reading, which was more of a further reading of public code than the Standard. I think we do want *some* Further reading later, but it should be content more related to the Standard.

Works towards #425. 

-----
[View rendered index.md](https://github.com/publiccodenet/standard/blob/ja-intro-refactoring/index.md)
[View rendered introduction.md](https://github.com/publiccodenet/standard/blob/ja-intro-refactoring/introduction.md)


ericherman:
I like the essence of this.

As was discussed in #425 and related community calls, the cut material is probably best suited outside.
Before merging this, would we like to see that content ported to the new location?
Or would we want to have this first and then re-create the related content elsewhere with our current understanding as a follow-on?

Ainali:
I think it would be wise to at least have a plan where it should go before we remove it. A short gap is probably not that bad, but if it turns into a longer period, it is a considerable amount of good background information that we are missing.

ericherman (approved):

ericherman:
Reviewing issue #218 makes me think that even the description of certification belongs outside the text of the standard and would fit better on publiccode.net

Ainali:
I suggest we have this pull request as a topic for the community call on Thursday, October 7.

nicorikken:
It saddens me to see the references to the larger context of Public Code and the reference to the Public Money? Public Code! campaign gone. Indeed this is a bold pull-request, and I can see this as the Standard for Public Code growing up and no longer needing these references in the standard. So I can see it in a positive light. As suggested by @ericherman I can see a place for this context on the website. I hope the Foundation for Public Code will keep explaining the values of Public Code and proven methods (to newcomers). As it is in the interest of the Foundation for Public Code to do so, I'm sure you will.

ericherman:
@nicorikken The content shouldn't be going away, rather moved from inside the Standard's introduction to on the publiccode.net websites. Outside of the intro, I expect this content to be expanded to suit multiple readers. I expect we won't even see this removal merged until similar material is merged elsewhere.

nebster (changes requested):
this definitely requires further thought about how much context about our work we can deliver with the standard.

Ainali:
Yes, I suggest that we close the pull request for now and start with something new when we have had time to come up with new ideas.

Ainali:
I had forgotten to link to the notes from the community call when this pull request were discussed: https://blog.publiccode.net/community%20call/2021/10/11/notes-from-community-call-7-october-2021.html

Ainali:
Closing this pull request for now as we have chosen a different direction and any work needed for that can start from a fresh branch.

PR #522, opened 2021-08-23
Simpler language in create reusable code

### LINKED ISSUE #389: Odd reference to "Public Code" in 'Create reusable and portable code'
This Public Code reference seems out of sync with how the Standard is currently used, especially since we [never agreed a labelling convention for mature codebases](https://github.com/publiccodenet/about/issues/223) stewarded by the Foundation for Public Code.

```
Codebases used by multiple parties have broad enough value to be Public Code and are more likely to benefit from a self-sustaining community.
```

Options:

- make it generic
- add a hyperlink or inline explanation

Ainali:
Fixes #389


ericherman (approved):
indeed, "multiple parties" and "broad enough value" is repeating the idea

Issue #456, opened 2021-02-22
Awkward sentence


ElenaFdR:
From the glossary:
```
 Revisions of code can be compared with each other and it makes it easy to restore previous versions.
```
It's not clear what the "it" in the 2nd half of this sentence refers to.

Ainali:
How about rewriting it to:

> Version control makes it possible to compare revisions of code with each other, and the visual difference between revisions helps when restoring a previous version.

Ainali:
Fixed by #525

PR #525, opened 2021-09-14
Issue #456 awkward sentence

ericherman:
Awkward sentence was highlighted in https://github.com/publiccodenet/standard/issues/456

### DIFF
--- a/glossary.md
+++ b/glossary.md
@@ -69,4 +69,7 @@ In revision (or version) control systems, a repository is a data structure which
 
 ## Version control
 
-Version control is the management of changes to source code and the files associated with it. Changes are usually identified by a code, termed the *revision number* (or similar). Each revision is associated with the time it was made and the person making the change thus making it easier to retrace the evolution of the code. Revisions of code can be compared with each other and it makes it easy to restore previous versions.
+Version control is the management of changes to source code and the files associated with it.
+Changes are usually identified by a code, termed the *revision number* (or similar).
+Each revision is associated with the time it was made and the person making the change thus making it easier to retrace the evolution of the code.
+Revision control systems can be used to compare different versions with each other and to see how content has changed over time.


glossary.md
```diff
@@ -61,4 +61,4 @@ In revision (or version) control systems, a repository is a data structure which
 
 ## Version control
 
-Version control is the management of changes to source code and the files associated with it. Changes are usually identified by a code, termed the *revision number* (or similar). Each revision is associated with the time it was made and the person making the change thus making it easier to retrace the evolution of the code. Revisions of code can be compared with each other and it makes it easy to restore previous versions.
+Version control is the management of changes to source code and the files associated with it. Changes are usually identified by a code, termed the *revision number* (or similar). Each revision is associated with the time it was made and the person making the change thus making it easier to retrace the evolution of the code. Revisions of code can be compared with each other and to see how content has changed over time.
```
Ainali, line 64:
```suggestion
Version control is the management of changes to source code and the files associated with it. Changes are usually identified by a code, termed the *revision number* (or similar). Each revision is associated with the time it was made and the person making the change thus making it easier to retrace the evolution of the code. Revisions of code can be compared with each other and makes it easy to see how content has changed over time.
```

Ainali: I think you lost this part when making the change.

ericherman: I think "easy" is not quite right for most audiences; I think the text could say "easier", as a relatively small portion of people describe git as "easy".

Ainali: I buy that, but we still need one more verb in this sentence.

Ainali: This is how I read the sentence: 

```
Revisions of code

  can be compared with each other 
and 
  to see how content has changed over time.
```
Skipping the first clause makes the sentence read as "Revisions of code to see how content has changed over time."

Ainali: I changed my proposal from "makes it easy" to "used" which means that it can be parsed as:

```
Revisions of code can be

  compared with each other 
and 
  used to see how content has changed over time.
```

which makes the sentence read coherently to me.

ericherman:
On re-reading, I think we still have an awkward sentence

> Revisions of code can be compared with each other and used to see how content has changed over time.

maybe:

> Revision control systems can be used to compare different versions with each other and to see how content has changed over time.

Ainali:
Yes, I like your latest suggestion, clearer language overall.

Ainali (approved):

Issue #512, opened 2021-07-26
Criteria navigation


jgroenen:
I would really love to be able to navigate the criteria for ease of reading.
![screenshot-standard publiccode net-2021 07 26-10_47_56](https://user-images.githubusercontent.com/1716809/126961262-5550bc8f-8af6-4ad0-bc74-08b27853ad61.png)

ericherman:
Is this issue different than https://github.com/publiccodenet/standard/issues/327 ?

Ainali:
This is a really nice idea. If it's not too hard to implement, I would rather want to see something like:

[« Welcome contributors](https://standard.publiccode.net/criteria/open-to-contributions.html) · [Maintain version control »](https://standard.publiccode.net/criteria/version-control-and-history.html)

ericherman:
Yes, I agree and I hope it wouldn't be too hard to do.

@Ainali I'll schedule some hack time for us to see if an hour-or-so of shaking our fists at the tools results in a PR that does this.

ericherman:
This is now live: https://standard.publiccode.net/criteria/code-in-the-open.html and notice the "next" link to " Bundle policy and source code →"

Issue #217, opened 2019-06-02
No navigation in web version


nomorepanic:
Documentation pages usually provide the index on each page for ease of navigation; some of the finer documentations introduce the next chapter at the end and provide a link for it.

The web version has neither and this means that a reader, once finished reading has to go back to the index.

bvhme:
This is absolutely right, and comes for now mainly from trying to keep the technical system simple to use. We should definetely update this to make it more usable.

ElenaFdR:
Feedback from Code for NL meetup:
Please expose nav when you're within a page so you can see (and jump to) what the other criteria are while you're reading.

ericherman:
See also #512

ericherman:
Resolved via #535

PR #535, opened 2021-11-11
Add prev/next criteria links (#217, #512)

ericherman:
See issues #217, #512

This adds previous and next links to each criteria page, but not the
index.

In _config.yml, the syntax which was previously used to ingore the index
page was incorrect, as it would match any page starting with 'i', 'n',
'd', 'e', or 'x'. Our attempts to craft the correct way to match all but
index in this regular expression dialect failed. For now, instead, we
match all pages except those which start with an 'i', recognizing that
this may create a problem in the future.


ericherman:
regarding the scope of the ignore, see https://github.com/publiccodenet/standard/issues/536

AlbaRoza (approved):
Go, go, go!

PR #545, opened 2021-11-29
MUST be in English, SHOULD be plain English

ericherman:
Upon re-reading it became evident that the SHOULD requirements
together comprise the concept of "plain English" for codebases
and the glossary definition needed to be exapanded or removed.

It did not make sense to put a MUST on "plain English" and a
SHOULD for all of the elements of "plain English". We believe
the spirit should clearly be a MUST for English and SHOULD be
as plain as possible.

-----
[View rendered criteria/understandable-english-first.md](https://github.com/ericherman/standard-for-public-code/blob/revert-plain-english/criteria/understandable-english-first.md)
[View rendered glossary.md](https://github.com/ericherman/standard-for-public-code/blob/revert-plain-english/glossary.md)

### DIFF
--- a/criteria/understandable-english-first.md
+++ b/criteria/understandable-english-first.md
@@ -6,7 +6,7 @@ order: 10
 
 ## Requirements
 
-* All codebase documentation MUST be in plain English.
+* All codebase documentation MUST be in English.
 * All code MUST be in English, except where required to execute as policy.
 * Any translation MUST be up to date with the English version and vice versa.
 * There SHOULD be no acronyms, abbreviations, puns or legal/non-English/domain specific terms in the codebase without an explanation preceding it or a link to an explanation.
--- a/glossary.md
+++ b/glossary.md
@@ -32,10 +32,6 @@ Open source is defined by the Open Source Initiative in their [Open Source Defin
 
 An open standard is any standard that meets the Open Source Initiative's [Open Standard Requirements](https://opensource.org/osr).
 
-## Plain English
-
-English on a lower secondary education reading level, as recommended by the [Web Content Accessibility Guidelines 2](https://www.w3.org/WAI/WCAG21/quickref/?showtechniques=315#readable).
-
 ## Policy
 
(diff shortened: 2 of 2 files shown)


Ainali (approved):

bvhme:
This does in my eyes take away a key argument for using English, mainly that it shouldn't just be English for English's sake but an English meant to connect to a larger group of people, not just people who speak English or are in your jargon.

It is my experience that when stated something needs to be English that people will directly think they need a native speaker to test the professionalism of something. This is diametrically opposed to what we're trying to achieve here (in my opinion). We're not trying to force codebases into the English cultural environment but hoping to use an international dialect of English meant to communicate widely so that more people can be included in communities and the local jargon is not employed.

To me 'English' without a qualifier means that the 'English' have no reason to double-check whether their words actually make sense to people from other cultural contexts. We know that native English speakers are just as bad at understanding they need to be clear in order to communicate across cultures. This qualifier served to make sure that even those who are native are held to a higher standard. In addition to this the plain communicates that organization specific jargon should not be used.

So in short, I believe removing 'plain' from the requirement undermines the goal of creating a framework that can work cross culturally and cross jargon.

Ainali:
> So in short, I believe removing 'plain' from the requirement undermines the goal of creating a framework that can work cross culturally and cross jargon.

I see your point, but we're not removing it from the requirement. It is not in the requirement today. With this PR we are just not adding it.

ericherman:
> it shouldn't just be English for English's sake but an English meant to connect to a larger group of people, not just people who speak English or are in your jargon.

I wholeheartedly agree! This thinking expresses exactly why we were hoping to add "plain" to this requirement of the criterion in this next revision of the Standard. Upon re-reading, @Ainali noticed what I have overlooked: that doing so in the "quick-and-easy" way sadly conflicts with the other requirements, thus we require a deeper change to the requirements of the criterion. The solution was to revert this bit for now, and I hope we shall revisit after the next release.

Issue #549, opened 2021-11-30
Policy Makers, Managers, and Developers and Designers lack glossary entry or definition


ericherman:
Readers' guide mentions a bit about what each should do, but not who is included in each group.

ericherman:
see also: https://github.com/publiccodenet/standard/issues/539

PR #550, opened 2021-11-30
Glossary for roles; issue #549

ericherman:
Co-authored-by:  Alba Roza <albaroza@gmail.com>

-----
[View rendered glossary.md](https://github.com/publiccodenet/standard/blob/glossary-issue-549/glossary.md)

### DIFF
--- a/glossary.md
+++ b/glossary.md
@@ -28,7 +28,7 @@ The public at large: end users of the code and the services based upon it.
 
 For example, a city's residents are considered end users of a city's services and of all code that powers these services.
 
-## Manager
+## Management
 
 People with decision making power or influence over the codebase and/or the developers and designers creating the codebase, including project managers, scrum masters, product owners, and community managers.
 
@@ -53,7 +53,7 @@ At the national level, policy and legislation (the law) are usually separate. Th
 
 In the Standard the word 'policy' refers to policy created and adopted by public organizations such as governments and municipalities.
 
-## Policymaker
+## Policy makers
 
 Politicians or senior civil servants that have decision making power or influence over what policy the codebase are aiming to implement.
 
--- a/glossary.md
+++ b/glossary.md
@@ -55,7 +55,7 @@ In the Standard the word 'policy' refers to policy created and adopted by public
 
 ## Policy makers
 
-Politicians or senior civil servants that have decision making power or influence over what policy the codebase are aiming to implement.
+Politicians or senior civil servants that have decision making power or influence over which policies the codebase is aiming to implement.
 
 ## Public code
 
--- a/glossary.md
+++ b/glossary.md
@@ -14,6 +14,10 @@ This can be – for example – a document or a version-control repository.
 
 In software engineering, continuous integration (CI) is the practice of merging all developers' working copies to a development branch of a codebase as frequently as reasonable.
 
+## Developers and designers
+
+Developers and designers includes anyone who is contributing to technical aspects of the codebase, including source code, templates, artwork, or any other asset that affect the execution of the running software, its tests, or its documentation.
+
 ## Different contexts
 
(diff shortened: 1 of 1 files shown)


ericherman:
@Ainali, I notice that "Manager" and "Policy maker" are singular where-as "Developers and designers" is plural. Is there a way to make "Developers and Designers" singular? Or make the others plural? Or is it best to leave as-is?

Ainali:
Good question! I think making the others plural is a good enough fix.

glossary.md
```diff
@@ -45,6 +53,10 @@ At the national level, policy and legislation (the law) are usually separate. Th
 
 In the Standard the word 'policy' refers to policy created and adopted by public organizations such as governments and municipalities.
 
+## Policy makers
+
+Politicians or senior civil servants that have decision making power or influence over which policies the codebase is aiming to implement.
```
bvhme, line 58: I describe this group as the people who decide what the outcomes should be of a given thing and its priority, as well as how this is codified in legal code.

***

Currently, this describes mostly a power relationship, not the roles these people have.
(Perhaps highlighting how we have less understanding of this target group)

***

My suggestion:

```suggestion
People tasked with formulating the outcomes, describing the goals, prioritising as well as developing the policies and legal code for a codebase.

For example:

* Senior civil servants
* Policy makers
* Policy advisors
* Politicians
* Political advisors
```

bvhme (changes requested):
This seriously took my brain for a loop. So it's a bit of a extensive contribution.

***

One of the interesting things here is that I wondered is: Does it add value to specify these precisely, or might these terms are more valuable being understood in a looser way and thus enabling more people to project them on to their own working environment.

I realize the earlier definition we wrote in here does not align with this thinking as well.

In addition to this, I think these group meaningfully overlap, and should meaningfully overlap, as a lot of problems, and roles in projects, aren't just fixed in these kind of boxes.
I fear that if we split them out precisely we are required to make sure that nothing falls through the cracks. 
For instance, if we explicitly mention 'the execution of the running software' does that mean that excludes the implementation of it?

For these reasons I've tried to start my feedback with what I would personally describe this group as.
And have added a few examples of people I might see in this category to each suggestion.
Following that I'll go in to trying to improve the descriptions introduced in the pull request, which might still be overly precise in my eyes.

My suggestions for these improvements, don't really replace my suggestion that it might be best to NOT define these groups at all, or maybe only give examples.

***

It could be that I'm seeing a problem that you are not, and then I'd love to hear about it.
Feel free to not take my advice for now and come back to this later.

But for now I'm wondering if these additions are an improvement, and thus I'm reviewing with 'Request changes'.

bvhme:
I might have been a bit too extreme this morning, I think defining these groups is fine, however I like the example roles as I think they might speak more than definitions, so I stand by my suggestions.

ericherman:
It's not about job title at all, it's about activity performed. For example, I expect the sections "Management: what you need to do" to be followed by a person with the title Senior Developer as this person performs management-like activities relating to the process, people, or outcomes.

Perhaps, instead of glossary entries (or perhaps in addition to glossary entries), the "Who this is for" section of the introduction should have more explanatory text. Key would be explaining the overlapping nature of the three areas of guidance as we have divided them.

Ainali:
I like your suggestion Eric. And the important part I think we wanted to solve with this PR was not a glossary entry in particular, but rather describing who these people are a bit more so that those that are affected feels included even if the title is not a one-to-one match to their current title.

bvhme:
Yes, good idea. Then it might be more about building, process and policy.

ericherman:
> it might be more about building, process and policy

I think this idea is important in the framing.

I'll give some thought to re-working this pull request into Introduction and possibly also glossary entries.

My inclination is to set this aside until next week, when I expect it to be quieter.

PR #552, opened 2021-12-01
Trigger rebuild

ericherman:
Merge develop back to main (no files changed)


Ainali (approved):
Confirm that no files are changed. Hopefully this will trigger a rebuild so that the new version number is picked up in the footer.

PR #572, opened 2022-02-03
Rephrase what developers need to do

### LINKED ISSUE #567: "reusable in different contexts" is not clear regarding using only part of a codebase
As @kerkuil highlights, there are times when there is value in reusing a part of a solution, rather than an entire solution.

The criterion "Create reusable and portable code" does not currently require modular design nor contain guidance encouraging modularity.

Ainali:
Less repetition and added that also parts of the source could be reused.

Fixes #567

### DIFF
--- a/criteria/reusable-and-portable-codebases.md
+++ b/criteria/reusable-and-portable-codebases.md
@@ -46,9 +46,11 @@ order: 3
 
 ## Developers and designers: what you need to do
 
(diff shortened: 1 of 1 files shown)


ericherman (approved):
Nice.

ericherman (approved):
Punctuation is a bit weird here, but not a blocker.

ericherman (approved):

PR #582, opened 2022-03-09
Replace BPMN SVG with Mermaid flowchart

### LINKED ISSUE #581: Convert BPMN diagram in the Standard with Mermaid
As we are not machine executing the process described in the Standard, the value of it being in BPMN is lower than the value of it being in a easily editable and translatable form.

Mermaid can be used to create a flowchart which would be as illustrative and easier to edit, especially for translations.

Ainali:
I am intentionally not deleting the SVG file yet. Do we want to keep it if someone external is linking to it? (I can't find any links from our repositories.)

Fixes #581

-----
[View rendered introduction.md](https://github.com/publiccodenet/standard/blob/mermaid/introduction.md)

### DIFF
--- a/introduction.md
+++ b/introduction.md
@@ -106,7 +106,28 @@ Every time a contribution is suggested to a codebase – through for instance a 
 
 The audit is presented as a review of the contribution. The codebase steward gives line by line feedback and compliance, helping the contributor to improve their contribution. The merge request cannot be fulfilled until the codebase stewards have approved the contribution.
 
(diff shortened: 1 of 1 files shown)


ericherman (approved):
Nice!

ericherman:
Leaving the SVG file for the moment will not harm anything, perhaps this is somethign to monitor? If we get no log requests for it after a while, we can delete it.

PR #603, opened 2022-03-29
Force size of badge image

### LINKED ISSUE #593: Move DPG logo in print version to be small and on the cover like the HTML version
Today, even though `width` and `height` are set in `print.html` the DPG logo in the print version is large and rendered on its own page when rendered by `weasyprint`.

It would be better to have the logo small and on the front cover, as in the HTML rendering.

Ainali:
Forced through the [--presentational-hints](https://doc.courtbouillon.org/weasyprint/latest/api_reference.html#supported-features?highlight=presentational) flag.

Moved the badge next to license image to fit on the front page.

Fixes #593

### DIFF
--- a/print.html
+++ b/print.html
@@ -279,9 +279,8 @@
     <img src="assets/hands-shake.svg" alt="hands shaking">
 
     <p>Version {{ site.version }}</p>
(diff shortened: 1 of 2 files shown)


ericherman (approved):
Rendered locally, looks good

PR #608, opened 2022-04-12
Add guide about language style

### LINKED ISSUE #586: Dash clause separator: "n-dash" vs. "m-dash" vs. "hyphen-minus"
In some places in the plain-ASCII dash (0x2D, in Unicode "hyphen-minus" U+002D) character is used as a clause separator.

However, in many places throughout the text, the Unicode EN DASH (U+2013) surrounded by spaces is used as a clause separator. 

American English style guides (e.g. Chicago Manual of Style, APA) recommend un-spaced EM DASHes (U+2014).

While I happen to think that un-spaced m-dashes look terrible compared to the (common in British English) variant of n-dashes with spaces, the rest of the text is in American English, thus we should either:

 * follow CMoS/APA
 * note this deviation from American English style in the CONTRIBUTING.md file
 * use ASCII-dash
 * make sentences more simple (rewrite uses of dash punctuation as clause separators)
 * something else?

Whatever we decide, adding a one-line in CONTRIBUTING seems like a good idea.

In the spirit of "Make contributing easy", there is something to be said for the less visually appealing "hyphen-minus".
In the spirit of "Use plain English", there is something to be said for more simple sentence construction. 

* https://en.wikipedia.org/wiki/Dash#En_dash_versus_em_dash

Ainali:
I am not sure if we want to document more closely the exact rules for dashes in the release guide or if this is enough.

Addresses the process part of #586

-----
[View rendered CONTRIBUTING.md](https://github.com/publiccodenet/standard/blob/contributing-style-guide/CONTRIBUTING.md)
[View rendered docs/releasing.md](https://github.com/publiccodenet/standard/blob/contributing-style-guide/docs/releasing.md)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,6 +29,12 @@ Add your changes in commits [with a message that explains them](https://robots.t
 
 If you are adding code, make sure you've added and updated the relevant documentation and tests before you submit your pull request. Make sure to write tests that show the behavior of the newly added or changed code.
 
+#### Style
+
+The Standard for Public Code aims to [use plain English](https://standard.publiccode.net/criteria/understandable-english-first.html) and we have chosen American English for spelling.
+However, we want to emphasize that it is more important that you make your contribution than worry about spelling and typography.
+We will help you get it right in our review process and we also have a separate quality check before [making a new release](docs/releasing.md).
+
 ### 2. Pull request
 
 When submitting the pull request, please accompany it with a description of the problem you are trying to solve and the issue numbers that this pull request fixes.
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -3,6 +3,7 @@
 1. Review state of the 'develop' branch
     - Ensure all changes intended for release are merged
     - Invite a proofread of the current state of the branch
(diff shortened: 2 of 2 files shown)


ericherman (commented):
This seems like a small improvement, for now.
It doesn't answer what the correct character is, but we can return to that in an additional commit.

Ainali:
> It doesn't answer what the correct character is, but we can return to that in an additional commit.

Yes, I couldn't figure out what we wanted from the issue, and that's why I kept this as a draft.

ericherman:
I'd like to revisit what @bvhme  wrote:
> If we do want consistency: I'd propose not dashing at all and guiding people to 'use simple sentences' and use commas or brackets if necessary.

With a brief skim through the text, I can think of equally clear ways to write the sentences I found without using dashes.

I think stronger guidance in the releasing could even be to rewrite sentences to replace dashes with commas or parentheses unless the use of dashes is clearly the best expression of the ideas.

If we find the exceptional case where we really wish to keep the dashes, I think we should follow CMoS/APA, and this would be to use un-spaced em-dashes.

ericherman (approved):

PR #613, opened 2022-04-25
Reorder the release guide, moving printing last

Ainali:
It feels more natural to complete the work in the repository before sending files to print. The numbers also imply that we cannot create the tag before printing, so putting it last prevents it from being seen as a blocker.

-----
[View rendered docs/releasing.md](https://github.com/publiccodenet/standard/blob/reorder-release-guide/docs/releasing.md)


ericherman (approved):
Naturally.

PR #617, opened 2022-05-17
Add link to community translations in footer

### LINKED ISSUE #615: Add link to community translations in the footer
Through publiccodenet/jekyll-theme#89 we have the possibility to add a link to the community translations in the footer. I suggest something like:

> [Community provided translations of the standard](https://publiccodenet.github.io/community-translations-standard/)

ericherman:
Fixes #615


Ainali (approved):

ElenaFdR:
(On a mobile, so the web interface is extra awkward): I suggest the hyperlink text should drop “provided”, because it doesn’t add detail that a person looking for a translation needs at the point that they’re looking. Presumably the provenance/limits of the translations are very clear once you get to the hyperlink destination.

By the same rationale, you could even drop “community”, because that could raise the question of where the non-community translations are.

ericherman:
We started with "community translations" then added "provided" to be more explicit.

ericherman:
And they are not yet at spot that they can act on it, thus I buy that angle.

Ainali (approved):
Yes, just "Translations" is better!

PR #627, opened 2022-06-14
update url-ignore regexes

ericherman:


ericherman:
https://github.com/publiccodenet/standard/runs/6873248154?check_suite_focus=true

Ainali (approved):
good generalization!

PR #629, opened 2022-06-15
Rename cibuild to test-all

ericherman:


ericherman:
hrm ... I see that the github diff view of "files changed" is not ideal for this, but if you "load diff" of the `cibuild.sh` file, you can see the total diff.

script/test-all.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+# This is script is provided for local development,
```
Ainali, line 3:
```suggestion
# This script is provided for local development,
```

Ainali: I believe this is a surplus "is".

Ainali (changes requested):

ericherman:
Extra "is" indeed, oddly that's been there for long, long time!

Ainali (approved):

Issue #264, opened 2019-09-05
propose a criterion for the Standard to ensure a codebase is easily "findable"


ericherman:
What's the minimum the Standard for Public Code should require of a codebase with regards to marketing? Is it something like, "MUST have a unique enough name that it is easily findable via a major search engine" or is it more sophisticated -- something about searching for functionality? How would we word the MUSTs, SHOULDs, and MAYs here?

Ainali:
This will be the topic of [the community call on June 2](https://hackmd.io/-OegeqvoThCbAsw3c3gIjw?view).

ElenaFdR:
What about encouragement to submit a codebase all potentially relevant software catalogs or indices?

Catalog examples:
- Inter-American Development Bank's [Code for Development](https://code.iadb.org/en/about)
- Italy's [national software catalog](https://developers.italia.it/en/software.html)
- the Intergovernmental Software Collaborative's [Intergovernmental Software Cooperatives](https://softwarecollaborative.org/cooperatives/)
- the Civic Tech Field Guide's [Tech list](https://directory.civictech.guide/tech)

ElenaFdR:
Ooh, something that didn't come up in our community call was mandatory publication of metrics (#265), a linked issue here. Shall we consider that as well in any potential new marketing criterion draft?

ericherman:
Regarding the connection to Open Data which was mentioned, the "Findable" aspect of the FAIR[1] Principles emphasizes metadata, more detailed than what "Create reusable and portable code" contains. Perhaps adding some detail would be useful.

[1] From [https://www.go-fair.org/fair-principles/](https://www.go-fair.org/fair-principles/):
> In 2016, the ‘[FAIR Guiding Principles for scientific data management and stewardship’] (http://www.nature.com/articles/sdata201618) were published in Scientific Data. The authors intended to provide guidelines to improve the Findability, Accessibility, Interoperability, and Reuse of digital assets.

ericherman:
As I've reflected on this, I see "Findability" and "Documentation" as better ways to frame requirements of public code codebases; "marketing" does not click as the right framing.

The essence of our talk was about "finding" a codebase, once found, there's all sorts of possibilities regarding presentation of various forms of documentation. Especially the "MUST" that we discussed: being able to find by descriptions of the problem to be solved without knowing the codebase name.

[...]

PR #630, opened 2022-06-20
Add new findability criterion

ElenaFdR:
Solves:
- #264 

This was discussed at the [June 2022 Standard for Public Code community call](https://blog.publiccode.net/community%20call/2022/06/08/notes-from-community-call-2-june-2022.html).

-----
[View rendered criteria/findability.md](https://github.com/publiccodenet/standard/blob/findability/criteria/findability.md)
[View rendered criteria/reusable-and-portable-codebases.md](https://github.com/publiccodenet/standard/blob/findability/criteria/reusable-and-portable-codebases.md)


Ainali:
Small general note that this does not yet align with how we use periods and capital letters in the other criteria.

criteria/findability.md
```diff
@@ -0,0 +1,51 @@
+# Findability
+
+## Requirements
+
+* The codebase MUST be findable using a search engine by description of the problem it solves in natural language
+* The codebase MUST be findable using a search engine by codebase name
+* Maintainers SHOULD submit the codebase to relevant software catalogs
+* The codebase SHOULD have a website which describes the problem the codebase solves using the preferred jargon of different roles or professions
```
Ainali, line 8: Should this perhaps end with something like "... that are users of the codebase"? This would narrow it down a bit and not suggest that all or a random set of roles and professions need to be catered for. Or is that implied enough through the context anyway?

criteria/findability.md
```diff
@@ -0,0 +1,51 @@
...
+* The codebase SHOULD have a unique and persistent identifier, such as a [Wikidata item](https://www.wikidata.org/wiki/Wikidata:Introduction) or [FSF software directory listing](https://directory.fsf.org/wiki/Main_Page), mentioning the major contributors, repository location and website
+* The codebase SHOULD include a machine-readable metadata description, for example in a publiccode.yml file
```
Ainali, line 10:
```suggestion
* The codebase SHOULD include a machine-readable metadata description, for example in a [publiccode.yml](https://github.com/publiccodeyml/publiccode.yml) file
```

Ainali: I believe the link got lost in the move from the previous criteria.

criteria/findability.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Developers and designers: what you need to do
+
+* search engine optimization
+* test your problem description with peers outside of your context who aren't familiar with the codebase
+* suggest conferences to present at
+
+## Further reading
+
+* The [FAIR Guiding Principles for scientific data management and stewardship](https://www.go-fair.org/fair-principles/) provide a nice list of attributes that make (meta)data more machine actionable (and hence more findable) - some of these apply directly to codebases, while others may provoke exploration into what the codebase equivalent would be
```
Ainali, line 51:
```suggestion
* The [FAIR Guiding Principles for scientific data management and stewardship](https://www.go-fair.org/fair-principles/) by the GO FAIR International Support and Coordination Office provide a nice list of attributes that make (meta)data more machine actionable (and hence more findable). Some of these apply directly to codebases, while others may provoke exploration into what the codebase equivalent would be.
```

Ainali:
@ElenaFdR Really good work turning the ideas into a full criterion! I only had a few minor comments, the essence of this is great.

criteria/findability.md
ElenaFdR, line 8:
```suggestion
* The codebase SHOULD have a website which describes the problem the codebase solves using the preferred jargon of different potential users of the codebase (including technologists, policy experts and managers)
```

ElenaFdR: take a look now?

criteria/findability.md
```diff
@@ -0,0 +1,51 @@
+# Findability
+
+## Requirements
+
+* The codebase MUST be findable using a search engine by description of the problem it solves in natural language.
+* The codebase MUST be findable using a search engine by codebase name.
+* Maintainers SHOULD submit the codebase to relevant software catalogs.
+* The codebase SHOULD have a website which describes the problem the codebase solves using the preferred jargon of different potential users of the codebase (including technologists, policy experts and managers.)
```
ElenaFdR, line 8:
```suggestion
* The codebase SHOULD have a website which describes the problem the codebase solves using the preferred jargon of different potential users of the codebase (including technologists, policy experts and managers).
```

criteria/findability.md
```diff
@@ -0,0 +1,51 @@
...
+
+* Budget for content design and Search Engine Optimization skills in the team.
+* Ensure people involved in the project present at relevant conferences.
+* Search trademark databases to avoid confusion or infringement before deciding the name.
+
+## Developers and designers: what you need to do
+
+* Search engine optimization.
+* Test your problem description with peers outside of your context who aren't familiar with the codebase.
+* Suggest conferences to present at.
```
ElenaFdR, line 47:
```suggestion
* Suggest conferences to present at and present at them.
```

ElenaFdR (commented):
This looks ready to me!

kerkuil:
I have mixed feelings about this. For me, there is a significant difference between developing codebased with accountability and reusability in mind and developing with actually governing the reuse itself. Which also implies things like community building and indeed findability. I would say the Standard should as best as it can scope itself to the source code in relation to the policy it serves. It should also facilitate the findability with proper indexable meta-data like with the publiccode.yml, but not describe how findability is actually reached.

kerkuil:
I believe this could be a 'bonus' feature. We discussed before that the standard is currently an all or nothing compliancy instrument. You either comply or you don't. I think it's still usefull to define categories in which certain requirements are grouped and to which you can comply or not. Or define a scale / priority to what you should comply first and what comes next. Findability as it is describe in this PR could then be a cherry on the cake. Where the requirement of just a publiccode.yml is more of a base requirement.

criteria/findability.md
```diff
@@ -0,0 +1,51 @@
+# Findability
+
+## Requirements
+
+* The codebase MUST be findable using a search engine by description of the problem it solves in natural language.
+* The codebase MUST be findable using a search engine by codebase name.
+* Maintainers SHOULD submit the codebase to relevant software catalogs.
```
sergei-maertens, line 7: I agree, but I at the moment there's quite a bit of chaos in NL about which software catalogs exist. It would be nice if the FFPC could provide resources of which catalogs may be relevant and even actively organize these catalogs.

criteria/findability.md
```diff
@@ -0,0 +1,51 @@
...
+* The codebase SHOULD have a website which describes the problem the codebase solves using the preferred jargon of different potential users of the codebase (including technologists, policy experts and managers).
```
sergei-maertens, line 8: With Open Zaak we now see that this website exists, but content-wise it's not really maintained nor clear who should maintain this. I do see the value in this, but I'm worried this will take away valuable development resources if you want to maintain this properly. So I'm not sure if this should be a SHOULD or MAY.

criteria/findability.md
```diff
@@ -0,0 +1,51 @@
...
+* The codebase SHOULD have a unique and persistent identifier, such as a [Wikidata item](https://www.wikidata.org/wiki/Wikidata:Introduction) or [FSF software directory listing](https://directory.fsf.org/wiki/Main_Page), mentioning the major contributors, repository location and website.
```
sergei-maertens, line 9: I'm not opposed to this, but at first glance from both websites it's not immediately clear to me what I have to do for this. So, this requires an investment for codebases to familiarize themselves with these concepts, and the benefit is not immediately clear to me why a codebase should have this.

criteria/findability.md
```diff
@@ -0,0 +1,51 @@
...
+* The codebase SHOULD include a machine-readable metadata description, for example in a [publiccode.yml](https://github.com/publiccodeyml/publiccode.yml) file.
+* There MAY be a dedicated domain name for the codebase.
```
sergei-maertens, line 11: How does this relate to having a website, which is a SHOULD? Wouldn't having a website imply that such a domain name exists?

criteria/findability.md
```diff
@@ -0,0 +1,51 @@
...
+* The community MAY present regularly at conferences.
+
+## Why this is important
+
+* Proactiveness is needed; just publishing a codebase and hoping it's found doesn't work.
+* The more findable a codebase is, the more potential new collaborators will find it.
```
sergei-maertens, line 17: Who should find the codebase? Practically in the case of Open Zaak I don't expect managers to go looking for `publiccode.yaml` on github for example. It seems that word of mouth and the VNG/GEMMA websites are the major drivers for findability.

criteria/findability.md
```diff
@@ -0,0 +1,51 @@
...
+
+* plain problem description
+* technical problem description
+* codebase name
+
+## Policy makers: what you need to do
+
+* Contribute a description of the policy area or problem this codebase acts on or operates.
+* Test your problem description with peers outside of your context who aren't familiar with the codebase.
+* Present on how the codebase implements the policy at relevent conferences.
```
sergei-maertens, line 35: Is this for the policy makers? The codebases implement policies made by policy makers, no? So I'd say this is downstream of policy makers and they are responsible for making it findable?

sergei-maertens (commented):
Some thoughts about the specific bullets.

I applaud the sentiment, but from my experiences so far I have doubts about the feasibility/priority/results of these changes. In particular the "taking at conferences" seems out of touch with the reality that people mostly want to take and use a codebase for free without investing back into the community.

jgroenen:
Some thought; If it is not findable, it is not reusable. But... there already is the "[Document codebase objectives](https://standard.publiccode.net/criteria/document-objectives.html)" and "[Code in the open](https://standard.publiccode.net/criteria/code-in-the-open.html)", which should cover the "find describing the problem it solves" and "find using the code base name".

I you go further and say: "create a (promotional) website, go out and talk about it" then I think there's some objective behind it that is different from the other criteria (which are about being (passively) open), not being mentioned; something like "actively attract users/collaborators" or "grow your community" or something like that. I'd try to formulate that first.

criteria/findability.md
Ainali: Yes, I agree that we should have a list of catalogs. That list might not quite fit in the catalog itself though, but perhaps a pointer to it. 

We are indeed planning a secondary resource where we provide a lot more examples of how the requirements in the standard could be met and possibly even linking to particularly elegant solutions.

criteria/findability.md
Ainali: That is a justified worry, especially for codebase communities that haven't grown to include several active public organizations yet (or doesn't have an active product steering team). As this happens, there is a natural tendency that not everyone can contribute through developers, but that their resources at hand might be in the communications or information departments. While they might not be very technical, they will probably be able to solve everyday tasks (of course dependent on the platform used).

criteria/findability.md
ericherman: @sergei-maertens consider simply having a repository well written README.md that serves the purpose, or other non-dedicated domain locations like example.org/software/foo and  example.org/software/bar ... these sites would allow a searcher to find the information, without a dedicated domain name.

criteria/findability.md
Ainali: I agree with you here that this is a bit opaque. I think we need to add a sentence in "Why this is important" so that the value becomes clearer. Perhaps something like:

> - By having well-written metadata and being part of the semantic web, it will be easier for other people to refer, cite, disambiguate and discover the codebase through third party tools.

And I see here that we again really could use a secondary resource with good examples of this. We should perhaps raise the priority for creating it.

criteria/findability.md
kerkuil: I also thinking the other way around. Software catalogs should only have to know where to find repositories for indexation. If we already have the `publiccode.yaml` criterion we don't need to enlist all repositories to catalogs but insteads should be able to only submit a (organizational) github, bitbucket, gitlab etc. account from where the catalogs can start there indexation. That makes findability a organizational responsability instead of a repository maintainer one.

criteria/findability.md
kerkuil: I don't believe this should be (solely) on a seperate website. I would like to see the information described in the repository itself. So you have one place that covers all necessary information to be able to use, implement and to asses the usefullnes of a solution.

criteria/findability.md
kerkuil: I would say this is a criterion for the catalog owners instead of the codebase maintainers.

criteria/findability.md
Ainali: In practice, and considering all types of codebases, I think all roles may go looking for codebases and need to find them. Yes, some codebases may be on the top-of-mind only for certain roles, and some roles are more likely to use a particular tool or method to find the codebase they need. For example, it might only be a handful of people looking for publiccode.yml file, and then it might be another, perhaps slightly larger handful, that looks in the catalogs that scrape these files whereas perhaps the majority will turn to a search engine (unless their country already have a prominent national catalog, like in Italy).

criteria/findability.md
kerkuil: There is no incentive for catalog owners to change there processes when we adopt this criterium. However, when we can create an apt like structure the incentive is clear.

criteria/findability.md
Ainali: We meant this for policy makers, yes. When policymakers (perhaps the CIO, or even the Mayor) of a municipality developing a codebase talk about it proudly, they will probably have an audience that is more likely to be able to make decisions to join the effort. So while the downstream also is important, we wanted to address an opportunity that is often overlooked.

criteria/findability.md
Ainali: You're correct, it shouldn't be solely on a separate website and indeed we already have requirements for this to be part of the documentation of the codebase. The documentation might be too detailed for findability and perhaps also not that easy to discover through search engines (and I don't think we necessarily want to create requirements that it should be). A landing page with a clear description with links to the repository and its documentation on its own domain or even on GH pages would be better suited for this task.

criteria/findability.md
ericherman: For example, a project to which I contributed has a gh-pages landing page:

https://openelectronicslab.github.io/OpenHardwareExG/

Which naturally links back to the repository itself, however there is no registered domain name like OpenHardwareExG.org or similar.

Ainali:
> Some thought; If it is not findable, it is not reusable. But... there already is the "[Document codebase objectives](https://standard.publiccode.net/criteria/document-objectives.html)" and "[Code in the open](https://standard.publiccode.net/criteria/code-in-the-open.html)", which should cover the "find describing the problem it solves" and "find using the code base name".
> 
> I you go further and say: "create a (promotional) website, go out and talk about it" then I think there's some objective behind it that is different from the other criteria (which are about being (passively) open), not being mentioned; something like "actively attract users/collaborators" or "grow your community" or something like that. I'd try to formulate that first.

Yes, that is a fair point. We have been careful about requirements that prescribe growth or are requiring continuous activity, since they are hard to state to be sustainable and testable. And there is nuance between encouraging only reuse versus collaboration. This criterion is intended to be in the spirit of the criterion [Make contributing easy](https://standard.publiccode.net/criteria/make-contributing-easy.html) and is chronologically earlier in the process for a new potential collaborator.

Ainali (approved):
Good to go!

ericherman (approved):

PR #637, opened 2022-06-28
Git ignore the "vendor/" directory

ericherman:
The vendor/ directory is typically where bundle installs gems


Ainali (approved):
Thank you!

PR #642, opened 2022-06-30
Add archive URLs step to releasing

### LINKED ISSUE #275: Submit external links to the Internet Archive
The pull request about archiving got me thinking about something related. We have several links to other websites in the standard and thus are a bit dependent on external parties. Would it make sense to, before each release of the standard, go through all external links and submit them to the [Wayback Machine](https://web.archive.org) so that they are available somewhere on the web even if the website we refer to goes down? We don't have that many, so this could probably be done manually (even though an automated way would be good to make sure not to miss any due to human error).

ericherman:
We run this by hand in order to avoid blocking merges to main on a very
slow script, given we can only submit 5 URLs per minute.

Fixes #275 

-----
[View rendered docs/releasing.md](https://github.com/ericherman/standard-for-public-code/blob/archive-urls/docs/releasing.md)

### DIFF
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -19,6 +19,7 @@
         - Reviewers will create issues for shortcomings found which would not prevent release
         - If needed for release, reviewers may create PRs to resolve issues
         - Re-request reviews if additional PRs are merged into release branch
+    - [ ] Run the to-archive-org.sh script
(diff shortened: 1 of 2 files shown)


Ainali (approved):
Nice future proofing!

PR #658, opened 2022-07-13
Pin html proofer to ~>3.19 to unbreak build

ericherman:
Newer html-proofer (4.0.1) fails with:

```
htmlproofer 4.0.1 | Error:  invalid option: --typhoeus-config
```

and complains about some of the jekyll generated html

https://github.com/publiccodenet/standard/actions/runs/2660759422

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,4 @@
 source "https://rubygems.org"
 gem 'github-pages', group: :jekyll_plugins
-gem 'html-proofer'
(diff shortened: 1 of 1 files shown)


ericherman:
to see the error locally, get a fresh copy of `develop` (or do a `git clean -dxff` in an existing checkout), then `bundle install` to get the latest html-proofer installed ... then `run script/test-all.sh`

PR #664, opened 2022-07-14
Add the "Draft" back to the version info on cover

ericherman:
fixes #660

### DIFF
--- a/print-cover.html
+++ b/print-cover.html
@@ -258,7 +258,7 @@
       <img src="assets/hands-shake.svg" alt="hands shaking">
 
       <div id="version-info">
-        <p>Version {{ site.version }}</p>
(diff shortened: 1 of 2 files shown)


Ainali:
This looks good on print.html, but on the PDFs the word 'Draft' is not turning into bold text when I generate them.

ericherman:
> This looks good on print.html, but on the PDFs the word 'Draft' is not turning into bold text when I generate them.

That is very strange because it renders bold for me; let's screen share and see if we can figure our what's different.

Ainali:
Yeah, the html's look fine, but I have WeasyPrint version 54.1, so that is probably it.

Ainali (approved):
Proven to work with the right setup, we should probably always do a visual check when generating PDFs to catch possible WeasyPrint oddities.

Issue #681, opened 2022-07-27
Rewrite the MAY in "Use open standards"


Ainali:
Perhaps we can rewrite the MAY in "Use open standards" to be a MUST NOT?

_Originally posted by @ericherman in https://github.com/publiccodenet/standard/issues/680#issuecomment-1196374283_

> Functionality using features from a standard that is not an open standard MAY be provided if necessary, but only in addition to compliant features.

Ainali:
Is this a possible phrasing?

> A codebase MUST NOT have functionality using features from a standard that is not an open standard unless it is only in addition to compliant features.

However, when thinking about this now, I think this is a kind of loophole (both the current wording and my suggested rephrasing). Provide one compliant feature, and everything else goes.

Instead, I think we need to add something like that these features can only be supplementary to the codebase and not core to its objective, but that will be much harder to write a requirement for.

ericherman:
Indeed, that can be read as a loophole. Okay, so let's consider the real-world use-cases.

An ISO standard, like the C programming language [ISO/IEC 9899:2018](https://iso-9899.info/wiki/The_Standard), does not meet the definition of an open standard, yet might be critically important for a codebase. The entire codebase may be covered by a non-open standard in this case.

It seems we want wording to achieve the smallest scope, yet the scope could remain rather large.

ericherman:
In the case of https://en.wikipedia.org/wiki/ISO_8601 this may be important for data exchange.

In both of these cases, it is not in reality a problem these standards are not open.

What are the problematic cases?

Let's review some of those and adjust the wording based upon what would actually hinder collaboration or reuse.

ericherman:
Perhaps something along the lines of:

> Use of a standard which is not open MUST NOT hinder collaboration or reuse.

Is that too loose?

Ainali:
> > Use of a standard which is not open MUST NOT hinder collaboration or reuse.

I think this is in the right direction!

ericherman:
I created draft pull request https://github.com/publiccodenet/standard/pull/685 as a possible rewording.

PR #685, opened 2022-08-01
Rewrite MAY as MUST NOT in "Use open standards"

ericherman:
Rather than state a requirement permissively when it is not desired,
change the phrasing to constrain/limit undesirable actions.

fixes: #681

-----
[View rendered criteria/open-standards.md](https://github.com/publiccodenet/standard/blob/open-standards-may-to-must-not/criteria/open-standards.md)


ericherman:
Alternative wording could be along the lines of:

* Any non-open standards chosen for use within the codebase MUST facilitate collaboration and reuse.

However, that may be too narrow.

ElenaFdR:
I find the proposal in the PR more clear than "Any non-open standards chosen for use within the codebase MUST facilitate collaboration and reuse", which could be misread as meaning standard marketing fluff ("our highly proprietary product increases the pace of your collaboration, letting you unlock valuable new synergies").

Ainali (approved):
Yes this is better! 

(We might need to be on the lookout for arguments like "paying for access to standards are not a hinder for a serious public organization" but those should be easy to spot.)

PR #688, opened 2022-09-06
Fix colliding links in pdf

Ainali:
The Wikidata and FSF links collides in the pdf, so move the introduction down to further reading which also works since it is a link to wider information than just about the item.

-----
[View rendered criteria/findability.md](https://github.com/publiccodenet/standard/blob/link-line-fix/criteria/findability.md)


ElenaFdR:
The current proposal makes it look like we endorse the FSF list over a Wikidata entry. I suggest removing the examples from the requirement, and putting them both instead into further reading or the implementation guide.

ericherman (approved):

PR #697, opened 2022-09-09
Update script/to-archive-org.sh from upstream

ericherman:
switches from wget to curl and puts standard output in a log file (by default in /tmp)


Ainali:
Does it make sense to add /tmp to .gitignore?

ericherman:
> Does it make sense to add /tmp to .gitignore?

If it were "./tmp" then yes, but /tmp is from the filesystem root and is (typically) mounted as a "tmpfs" meaning that the data does not survive reboots.

Ainali (approved):

PR #706, opened 2022-09-19
Label maturity as beta instead of draft

### LINKED ISSUE #704: Add "beta" label to align with Document Codebase Maturity

Ainali:
Fixes #704

Added extra documentation so that it also shows on the web and in the repository.

-----
[View rendered README.md](https://github.com/publiccodenet/standard/blob/draft-to-beta/README.md)
[View rendered docs/standard-for-public-code.md](https://github.com/publiccodenet/standard/blob/draft-to-beta/docs/standard-for-public-code.md)
[View rendered index.md](https://github.com/publiccodenet/standard/blob/draft-to-beta/index.md)

### DIFF
--- a/print-cover.html
+++ b/print-cover.html
@@ -253,7 +253,7 @@
       <img src="assets/hands-shake.svg" alt="hands shaking">
 
       <div id="version-info">
-        <p><strong>Draft</strong><br/>Version {{ site.version }}</p>
+        <p><strong>Beta</strong><br/>Version {{ site.version }}</p>
         <p><img src="/assets/cc-zero-badge.svg" alt="Licensed CC0" height="46"> <img src="/assets/dpg-badge.svg" alt="Digital Public Goods approval badge" width="92" height="46"></p>
         <p>https://standard.publiccode.net/</p>
       </div>
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ The standard lives at [standard.publiccode.net](https://standard.publiccode.net/
 
 ## Help improve this standard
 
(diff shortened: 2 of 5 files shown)


print-cover.html
```diff
@@ -253,7 +253,7 @@ <h1 class="title">Standard for Public Code</h1>
       <img src="assets/hands-shake.svg" alt="hands shaking">
 
       <div id="version-info">
-        <p><strong>Draft</strong><br/>Version {{ site.version }}</p>
+        <p><strong>Beta</strong><br/>Version {{ site.version }}</p>
```
ericherman, line 256: In the context of a document, the word "Draft" is more meaningful than "Beta" for most readers.
As such, I'd suggest we continue to say "Draft" in bold and perhaps add "beta" after the version.

ericherman (changes requested):

Ainali:
_This comment is partly a summary of a discussion @ericherman and I had this morning._
I should have noted that the footers pulls in "pre-release" from GitHub, which is another status we mention in the Standard. However, [GitHub documentation](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository) states "To notify users that the release is not ready for production and may be unstable" which corresponds better to our definitions of prototype and alpha.

That poses the question if we should uncheck that box? And if we do, do we want to be more explicit on the [release page](https://github.com/publiccodenet/standard/releases) and how would we do that? 

A common way to do it with [Semantic versioning](https://semver.org/) is to append something to the end of the version number, for example **0.4.0-beta**.  If we used that, we wouldn't need to repeat it in text in other places.
After some thinking, I like this solution more and more.

Ainali:
Superseded by #729

PR #723, opened 2022-10-12
Replace unreliable omg.org link

ericherman:
As https://www.omg.org/ has not been reliable, this converts the links to wikipedia.org articles about the subjects.

See also 01eb8551b4001802129dc8eb85b7f08615a83afc

-----
[View rendered criteria/bundle-policy-and-code.md](https://github.com/publiccodenet/standard/blob/omg-to-wikipedia/criteria/bundle-policy-and-code.md)


Ainali (approved):
Nice!

Since we link to these Wikipedia articles now, it does make sense to give a hand and improve the CMMN one so that the hat note quality disclaimer can be removed.

ericherman:
> Nice!
> 
> Since we link to these Wikipedia articles now, it does make sense to give a hand and improve the CMMN one so that the hat note quality disclaimer can be removed.

I think that would be a good way for us tocontribute, yes.

PR #736, opened 2022-11-07
Add list of Standards

### LINKED ISSUE #718: Make list of chosen standards for the Standard
In [our review](https://github.com/publiccodenet/standard/blob/develop/docs/standard-for-public-code.md) of:

> Any standard chosen for use within the codebase MUST be listed in the documentation with a link to where it is available.

We noted that "RFC2119 is referenced in the reader's guide, need to make a complete list"

Let's compile a list and add to CONTRIBUTING.

Ainali:
Fixes #718

Start this as a draft because I am uncertain if this is the only standard we apply or if there are more to be added.

-----
[View rendered CONTRIBUTING.md](https://github.com/publiccodenet/standard/blob/list-of-standards-used/CONTRIBUTING.md)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,6 +48,14 @@ The Standard for Public Code aims to [use plain English](https://standard.public
 However, we want to emphasize that it is more important that you make your contribution than worry about spelling and typography.
 We will help you get it right in our review process and we also have a separate quality check before [making a new release](docs/releasing.md).
 
+#### Standards to follow
+
+These are the standards that the Standard for Public Code uses.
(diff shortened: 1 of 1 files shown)


ericherman (approved):
We should merge this as-is.
I think we could add more, e.g.: https://www.w3.org/TR/WCAG21/#readable

Ainali:
> We should merge this as-is. I think we could add more, e.g.: https://www.w3.org/TR/WCAG21/#readable

I added that now. I think we can try to be as complete as possible in order to say that #718 really gets fixed.

ericherman (approved):

PR #751, opened 2022-11-15
Improve glossary "respository" for relevance

### LINKED ISSUE #369: Glossary items are not uniform
Right now glossary items are all different, some contain examples and others have source references.

I propose giving these all the same form: Our definition of what something is with optional further reading links. I propose not having sources as I think we should own the definitions in our glossary.

### LINKED ISSUE #394: Glossary definition for 'repository' has unhyperlinked source
The source appears to be [SVNBook](http://svnbook.red-bean.com/), but it's not clear which version. Our sentence isn't a straight quotation.

ericherman:
fixes #369
fixes #394 

Co-authored-by: @Ainali

-----
[View rendered glossary.md](https://github.com/publiccodenet/standard/blob/glossary-repo/glossary.md)

### DIFF
--- a/glossary.md
+++ b/glossary.md
@@ -64,7 +64,9 @@ To facilitate reuse, public code should be either released into the public domai
 
 ## Repository
 
-In revision (or version) control systems, a repository is a data structure which stores metadata for a set of files or directory structure. (source: SVNBook)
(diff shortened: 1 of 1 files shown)


Ainali (approved):
I think this is a good improvement!

PR #756, opened 2022-11-21
Modify SVG to become more translation friendly

Ainali:
- The SVG has been un-minimized for better overall readability
- All text has been moved closed to the top
- Backgrounds have been removed to make the above simpler code (also not needed on our sites or in print)
- Long texts use [flow* tags](https://www.w3.org/TR/2004/WD-SVG12-20041027/flow.html) so that translators do not need to consider line wrapping

**However**, browsers seem **to not** have support for this, so a conversion to PNG is probably the way forward


Ainali:
I am not sure if Weasyprint can create PNGs, but Inkscape can do it with this command:

`inkscape images/audit-flow.svg -o images/audit-flow.png`

ericherman:
Superseded by #761

PR #765, opened 2022-11-23
Change version badge from red to yellow

ericherman:
Yellow is commonly a warning color, where as red is commonly an error color. We wish to warn that this is a draft, subject to change, but it is a pre-release that can be tested today, more like a release candidate than an unstable alpha, thus yellow more closely communicates that level of warning.

-----
[View rendered README.md](https://github.com/publiccodenet/standard/blob/yellow-badge-of-courage/README.md)


Ainali (approved):
Good point!

PR #768, opened 2022-11-28
Ping translation contirbutors after release

ericherman:
null

-----
[View rendered docs/releasing.md](https://github.com/publiccodenet/standard/blob/ping-trans/docs/releasing.md)


Ainali:
We already have it on https://about.publiccode.net/activities/releasing-the-standard/#after-the-release
Do we want it in both places?

ericherman:
I feel like we work from the releasing doc directly, thus it's more useful there than in about's activities.
(I do not mind it being in both places.)

Ainali (approved):

PR #779, opened 2022-12-01
README: Add deploy badge

ericherman:


Ainali (approved):
It works, but should the other badges also have `/actions/` in the URL?

PR #787, opened 2022-12-06
use jekyll to generate license.html

ericherman:


ericherman:
@Ainali the sidebar created by `page-information` makes the text horizontal scroll.
We do not want that, but I do not yet know how to get rid of it.

Ainali (approved):
The inline CSS looks good. And it also works in the print version!

PR #792, opened 2022-12-07
Split changelog

ericherman:
Co-authored-by: @Ainali


ericherman:
We'll need to contend with the fragments not having a top-level header:

```
changelog/changelog-0.1.0.md:4: MD002 First header should be a top level header
```

Perhaps there is a form of `include` which will down-shift the headers by one level?

Ainali:
> We'll need to contend with the fragments not having a top-level header:

Or we could exclude that entire directory from the build process in `_config.yml`.

Ainali (approved):

PR #793, opened 2022-12-08
move mdl exceptions from script to .mdlrc/.mdl_style.rb

ericherman:
This opens the door for greater control over mdl.

### DIFF
--- a/CC0-1.0.md
+++ b/CC0-1.0.md
@@ -5,16 +5,16 @@ redirect_from:
     - LICENSE.html
 ---
 
+# Creative Commons Zero v1.0 Universal
+
 <!-- SPDX-License-Identifier: CC0-1.0 -->
 <!-- SPDX-FileCopyrightText: 2022 The Foundation for Public Code <info@publiccode.net>, https://standard.publiccode.net/AUTHORS -->
 
 <!-- TODO FIXME: move styling to the proper place, not inline html -->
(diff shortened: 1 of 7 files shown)


CC0-1.0.md
```diff
@@ -5,16 +5,16 @@ redirect_from:
...
 
 <!-- TODO FIXME: move styling to the proper place, not inline html -->
 <style>.highlight { flex: 2 2 80%; max-width: 100%; } </style>
 
-# Creative Commons Zero v1.0 Universal
-
 This license is the legal contract that allows anyone to do anything they like with the content in this entire document.
 
-```
+```text
```
Ainali, line 18: Was this "text" intentional?

ericherman: Yes, `MD040 Fenced code blocks should have a language specified`

ericherman: It's not a new rule (as there are 47), but it only just started triggering locally for me, possibly the result of a recent new bundle install.

ericherman: Interesting. Testing reveals that there is a difference in `mdl` results when using the commandline to exclude rules and when using these `.mdlrc` and `.mdl_style.rb` .... results are less "fussy" when using the commandline exclusions. At a guess, perhaps it is because the default from the commandline is less inclusive than the `all` that is at the top of `.mdl_style.rb`? Regardless, I think "more fussy" is better than "less fussy" for our case here.

PR #800, opened 2022-12-21
Only run linkcheck on publiccodenet/standard

### LINKED ISSUE #799: Scheduled link check should not run on forks
So far, contributors have created forks of the repository in order to contribute, not to host an alternative instance. As such, only the [Test](https://github.com/publiccodenet/standard/actions/workflows/test.yml) workflow makes sense. The other workflows, especially the [Scheduled link check](https://github.com/publiccodenet/standard/actions/workflows/link-check.yml), should not run on forks.

ericherman:
Thanks to @rgs and @niven!

Only run the Scheduled link check's "cibuild" if the repository is `publiccodenet/standard`, so that forks do not have to deal with failing jobs.

fixes #799

### DIFF
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -21,6 +21,7 @@ defaults:
 
 jobs:
   cibuild:
+    if: ${{ github.repository == "publiccodenet/standard" }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


Ainali:
Hm, [it looks like](https://github.com/Ainali/standard-public/actions/runs/3826340629) it tries to run but fail, rather than skipping. Or am I just reading it wrong?

rgs:
You're right. That's because github actions only understand single-quoted strings. <https://docs.github.com/en/actions/learn-github-actions/expressions>

Ainali:
Yes, with single quotes, it [very nicely says that it was skipped](https://github.com/Ainali/standard-public/actions/runs/3834616671).

Ainali (approved):

PR #803, opened 2023-01-05
Add applying the standard section to README

### LINKED ISSUE #309: Do we need to have a philosophy for 'breaking changes'?
A thought struck me during a conversation about another open standard and that is how to deal with certifications. In our case that would be mature codebases. What happens when we introduce new requirements that potentially make a previous compliant codebase not be compliant anymore? Should we have a philosophy of announcing such changes with enough time to give them a chance to stay compliant? Or is this so hypothetical that we can leave this for now?

Ainali:
While we had good sections on contributing to the standard, we were missing a section on how to use it.

Also fixes #309

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,19 @@ The Standard for Public Code is in a draft format.
 We are preparing it for a version 1.0 release.
 Currently, we are testing it on a small number of codebases.
 
+## Applying the Standard for Public Code on your codebase
+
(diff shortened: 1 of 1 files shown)


ericherman (approved):

ericherman:
I expect we shall refine the language a bit, but I suggest we merge this as-is, and tweak the language in follow-ups.

Issue #213, opened 2019-05-21
improving categories for recipients of the standard


MirjamvT:
Current categories (throughout the standard) are 
- policy makers
- management
- designers and developers
I suggest to at least make it grammatically consistent (management > managers)
but 
I would also prefer it if we can be more specific, maybe writing 'public administrators' (which would include e.g. people in procurement), but this would erroneously exclude people like business analysts, (commercial) providers of services).
Suggestions welcome!

PR #810, opened 2023-01-10
Rename audiance from "Managment" to "Managers"

ericherman:
See issue #213


Ainali (approved):
Looks good! We should probably take a look in our other repositories too in how we refer to them there, but that is a follow-up, not a blocker at all.

Issue #560, opened 2021-12-14
Idea: 'why is this important' before requirements


bvhme:
The standard might become more readably for a wider audience if the criteria don't start with the, often tough and technical, requirements. Arguably, the 'why this is important' is the most important.

I'd propose putting the 'why this is important' before the requirements. This will make it a bit tougher to use the standard as a tool but I expect will vastly improve the speed someone can get acquainted with it.

Ainali:
I like your reasoning for this. It's a quite large change though since most of the bullet points are written in the context of someone who have just read the requirements. We'll have to some more editing to them for them to make sense if we put them on top.

ericherman:
@Ainali Yes, in order for this to be coherent, this is more than a cut-and-paste, however it should be something that the two of us can knock-out in a few working sessions. I suggest we create a branch for this and work on it incrementally. Shooting from the him, I expect we can find time to work through them all before end-of-january.

ericherman:
It's worth reviewing #229 and #230 as we prioritize "Why" in the criteria layout.

Ainali:
When doing this, some inspiration might be the videos we created for [the Standard course](https://citybycity.academy/course/standards-for-smart-cities) as they were an introduction that came before the requirements.

ericherman:
When creating the first draft ( #811 ), many of the ideas expressed in "What this does not do" would naturally fit in the preambles. Would it make sense to hoist those up as well?

#813

PR #811, opened 2023-01-11
Move "Why this is important" to before requirements

ericherman:
fixes: #560

### DIFF
--- a/readers-guide.md
+++ b/readers-guide.md
@@ -12,6 +12,10 @@ It is common for individuals to have duties which span multiple roles.
 
 Below is a brief explanation of each of these sections and how they are used within the criteria of the Standard.
 
+## Introduction
+
+This section explains what the criterion aims to achieve and why it is important for a codebase's users and contributors.
+
 ## Requirements
 
 This section lists what needs to be done in order to comply with the standard.
@@ -28,10 +32,6 @@ The following keywords in this document are to be interpreted as described in [I
 * RECOMMENDED
 * MAY
 * OPTIONAL
-
-## Why this is important
-
-This section explains why it is important for the users and contributors of this codebase that these requirements are followed.
 
 ## What this does not do
 
--- a/criteria/bundle-policy-and-code.md
+++ b/criteria/bundle-policy-and-code.md
@@ -5,18 +5,16 @@ order: 2
 ---
 # Bundle policy and source code
 
+Access to both source code and policy documentation provides building blocks for anyone to implement the codebase in their local context or  contribute to the further development of the codebase.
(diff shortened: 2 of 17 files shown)


Ainali:
Very nice and good to get started on this! I was looking at the videos we made for our course earlier and inspired by that I combined it with your suggestion for Bundle policy and code to this:

> When someone wants to collaborate in, or reuse, a codebase they need contextual information about the purpose and objective the codebase to be able to evaluate whether to adopt that codebase in their context. To get that level of understanding it is necessary that they have access to both the source code and the policy documents. That access is fundamental to understanding what problems a codebase is trying to solve and how it sets out to solve them. It is also required in order to be able to understand what process changes an organization must choose to make or how to contribute additional configurability to the existing solution in order to adapt it to the new context.
> 
> Therefore, providing access to both source code and policy documentation provides building blocks for others to reimplement the codebase in their local context or as a starting point to contribute to the development of the codebase.

I am not sure that this is a clearly different style, or if it is just a bit reordered and longer text, but at least to me, it reads a bit smoother.

Ainali:
To recap some discussion that we have had this morning, these new lead section should be less of a direct answer to the question "Why is this important" but rather be something of a rationale for the entire criterion (almost like the recitals in EU directives or pre-ambles in other legislative documents).
Therefore, we should check that all requirements are possible to derive from the lead section. If not, it might suggest that we need to add more to the pre-amble, or that the requirement really fits better in another criterion. (It's possible, but less likely, that this will also reveal some requirements that are really beyond the scope of the standard.)

ericherman:
This branch now contains a rough draft of what it might look like.

Many of the sentences need work, but we hope the ideas are there.

We have identified several issues which suggest that some requirements might fit better under different criteria:
* #814
* #815
* #817
* #819
* #820

Ainali (approved):

ericherman:
https://blog.publiccode.net/community%20call/2023/02/02/notes-from-community-call-3-february-2023.html

readers-guide.md
```diff
@@ -11,6 +11,10 @@ It is common for individuals to have duties which span multiple roles.
 
 Below is a brief explanation of each of these sections and how they are used within the criteria of the Standard.
 
+## Preamble
```
ElenaFdR, line 14:
```suggestion
## Introduction
```

'Preamble' is a less common word than 'introduction', and carries more (unnecessary) overtones of legal jargon.

Ainali:
I noticed that in many places it is no longer the first time a term is used that ew are linking to the glossary. I think we need to look through each criterion and see if the glossary links are in the right place.

ericherman:
Ah yes, quite right. Let's do a pass at that.

Ainali (approved):

PR #826, opened 2023-01-19
Clarify style guide to specify expectations

### LINKED ISSUE #820: "inline comments for non-trivial sections" should move from Style to Document Code

ericherman:
fixes #820

Co-authored-by: Jan Ainali <jan@publiccode.net>


Ainali (approved):
I like this change!

Note that this might be a breaking change for codebases that have a good habit of making inline comments, but where that is not explicit in any guidelines.

PR #827, opened 2023-01-19
Clarify that machine readable closed is not better than open

### LINKED ISSUE #825: Clarify that machine testable should not be reason to chose a close standard over an open one.

ericherman:
fixes #825

Co-authored-by: Jan Ainali <jan@publiccode.net>


Ainali (approved):
Yes, that is now totally unambiguous!

PR #831, opened 2023-01-19
WIP: exploring qr codes for our links

ericherman:
This is a work in progress to explore adding QR codes.

For now, it can be seen by using `script/serve.sh` and going to the http://127.0.0.1:4000/docs/qr-links.html page.

We must explore what could make sense for the print version.


ericherman:
This branch served as a useful exploration of creating QR codes.

We have not answered the question of how to integrate the codes automatically.

We have [updated the Longer term roadmap](https://github.com/publiccodenet/standard/pull/878) to include adding QR codes for external links in print version.

PR #848, opened 2023-02-13
Changes to increase findability of the Standard

ElenaFdR:
To help the Standard meet the Standard better.


ericherman (approved):

Ainali (approved):

ericherman:
I see this is draft; I think draft is already an improvement, thus I'd suggest that we could merge these changes and continue to iterate in future increments.

PR #851, opened 2023-02-20
Less focus on contributor behavior

### LINKED ISSUE #846: Many requirements say "contributors SHOULD" or similar - review for better wording
Rather than talk about what people do, we can talk about what expectations are described in the codebase documentation. 

For instance, as a contributor I might initially not group relevant commits together, in this case, a maintainer may decide to do the re-grouping as a part of their work in shepherding the patch into the main development branch rather than pedantically enforce that I, the contributor, must be the one to fix the patch.

ericherman:
fixes #846

### DIFF
--- a/criteria/code-in-the-open.md
+++ b/criteria/code-in-the-open.md
@@ -12,7 +12,7 @@ Together, this creates more opportunities for citizens to understand how softwar
 
 * All source code for any policy in use (unless used for fraud detection) MUST be published and publicly accessible.
 * All source code for any software in use (unless used for fraud detection) MUST be published and publicly accessible.
-* Contributors MUST NOT upload sensitive information regarding users, their organization or third parties to the [repository](../glossary.md#repository).
+* The codebase MUST NOT contain sensitive information regarding users, their organization or third parties.
 * Any source code not currently in use (such as new versions, proposals or older versions) SHOULD be published.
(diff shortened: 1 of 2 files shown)


Ainali:
This is really an improvement!

I wonder if the "Contribution review MUST verify that no sensitive information is being contributed." could be rephrased so that it doesn't focus on the review itself, as we almost need to be there and see it to verify.
Is it too roundabout to also talk about guidelines here?

ericherman:
I had a very similar thought, but in other places we talk about reviewers, too.

Maybe we can be prescriptive about reviewer behaviour, but I lean in the direction of thinking we should talk about review policy/guidelines as well.

ericherman:
I compromised with "review must include" as being non-specific about which reviewer does what, but that somewhere along the process it happens. I'm not sure it's a clear enough compromise.

Ainali (approved):

Ainali:
We can perhaps improve it even further later on, but just as it is, it is still better than what we have.

ericherman:
Okay, I propose we let's let this sit a couple of days at least; maybe we will have more ideas or get more input.

Ainali:
Thinking about it, since the first of the requirements says, "MUST NOT contain" it is really implied that it has to be reviewed for it because if such things are merged, the first requirement fails. So perhaps the requirement about the review is not really needed.

Ainali (approved):

PR #856, opened 2023-02-22
Add sitemap

Ainali:
Inspired by #852 here is a Jekyll plugin for generating a sitemap.


ericherman (approved):
http://127.0.0.1:4000/sitemap.xml looks good

PR #861, opened 2023-02-22
Plain english: prefer "make sure" over "ensure"

### LINKED ISSUE #860: Plain English: "ensure" versus "make sure"
I believe that "make sure" is more plain English than "ensure".

Should we take a close look at the entire Standard and ~ensure~ make sure that we only use the simpler of the two terms in all places?

ericherman:
fixes #860

### DIFF
--- a/criteria/bundle-policy-and-code.md
+++ b/criteria/bundle-policy-and-code.md
@@ -30,7 +30,7 @@ To be able to evaluate whether to implement a codebase in a new context, an orga
 
 ## Policy makers: what you need to do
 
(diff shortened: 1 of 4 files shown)


Ainali (approved):

ericherman:
There are a couple of places where "ensure" is still in use, but seem to me to be better in those situations.

```
git grep -i 'ensure' criteria/*.md introduction.md
introduction.md:Local responsibility and democratic accountability are ensured when a public organization implements and maintains their own public code.
introduction.md:The Foundation for Public Code ensures that codebases under its stewardship (and not in incubation or the attic) are compliant with the Standard for Public Code.
```

PR #866, opened 2023-02-23
Remove certification process

### LINKED ISSUE #396: Remove sections on how the Foundation uses the Standard from the introduction
Replace these with a statement of the Standard's importance to our mission and practice, add a hyperlink to About or marketing materials for what we offer and how we use the Standard.

Why: 
- the current text was written before we had codebases in incubation, and doesn't reflect what we know now about codebase stewardship
- the Standard will be easier to use by a wider range of actors (not all of whom may be interested in our stewardship) without these paragraphs
- removing our obsolete process description from the Standard reduces future content debt - it means we have fewer, better focused places to keep our process description accurate, and allows the Standard intro to act primarily as rhetoric (a call to arms)
- making this text shorter but better makes translation/maintaining translation easier and cheaper

ericherman:
The (untested) certification process that we plan to employ as part of codebase stewardship is duplicated on the publiccode.net website, where it fits better.

https://publiccode.net/standard-for-public-code/

The text there has also been improved, and has diverged from this text slightly.

fixes #396


Ainali:
I think this fixes #396.

Ainali (dismissed):

bvhme (approved):
This seems like a good idea.

Ainali (approved):

ElenaFdR (approved):
Neat and tidy!

PR #873, opened 2023-03-02
Criteria list should be generated, not enumerated by hand

ericherman:
This generates the list programmatically, rather than having them listed by hand, thus ensuring the same ordering.

This commit does address that in other places, we have the criteria as a numbered, rather than bulleted list.


Ainali (approved):
Elegant!

PR #875, opened 2023-03-02
Allow navigation back to critera from first criterion

### LINKED ISSUE #536: Pattern in _config.yml to exclude criteria/index is too broad
In 45f6619538129dd80404b843aeb05a6da9c60418 it was discovered that the regex to match all criteria pages, excluding the index page was incorrect. It was replaced by an overly-broad expression which matches all criteria pages except those beginning with an 'i'. 

A better expression would match exactly all pages except the 'index' page.

ericherman:
And also to first criterion from criteria via next link.

fixes #536

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -25,8 +25,6 @@ include_translations: true
 defaults:
   -
     scope:
-      # ignore criteria/index.html, but match all other criteria pages
-      # (match all criteria pages not starting with an 'i')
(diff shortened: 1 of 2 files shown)


Ainali (approved):

ericherman:
If we wish, can toggle-off the navigation via setting `show_prev_next: false` in criteria/index.md frontmatter.

PR #879, opened 2023-03-08
script/generate-review-template: fix file list

ericherman:
As of 96202039be524a83b5597f31eb67c843820816d8 the list is no longer enumerated by hand, thus this script must also generate the list programmatically.


Ainali (changes requested):
When I run it locally, it puts a SPDX thing in front of every criterion.

Ainali (approved):

PR #880, opened 2023-03-08
unlink glossary terms in review template

ericherman:
Since b60c9c4 fewer of the criteria have glossary terms linked. The remaining ones now are out-of-place.
This unlinks those pages for now.
If we want links, we should ensure the first usage of each term is linked.

### DIFF
--- a/docs/review-template.md
+++ b/docs/review-template.md
@@ -11,45 +11,47 @@ Link to commitment to meet the Standard for Public Code:
 
 Requirement | meets | &nbsp;links&nbsp;and&nbsp;notes&nbsp;
 -----|-----|-----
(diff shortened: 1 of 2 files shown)


Ainali (changes requested):
Still seem to generate some glossary links. For example in Code in the open I get 

```
Documenting which source code or policy underpins any specific interaction the [general public](../glossary.md#general-public) may have with an organization is OPTIONAL. |  |
```

ericherman:
> Still seem to generate some glossary links.

I'm not able to reproduce that error. I've updated the result as a separate commit.

Ainali (approved):
I can't reproduce it either anymore. I wonder if I briefly was in a weird state.

ericherman:
Okay, I'm glad that it's not just "it works on my box" ... I suggest we merge as-is.

PR #900, opened 2023-04-11
Lower significance of search engine requirements

### LINKED ISSUE #896: Focus on marketing instead of the problem
In [Make the codebase findable](https://standard.publiccode.net/criteria/findability.html), we say:

> The codebase MUST be findable using a search engine by describing the problem it solves in natural language.

However, for many codebases, it is hard to know exactly what the problem it solves is or the type of codebase is already so well-known that any search for it will include a mention of the type. Perhaps this should instead focus on that the codebase should be findable through the way it markets itself, rather than guessing how others will try to find it.

Ainali:
As the results of the search engines are largely out of control for a codebase, these requirements need lower significance.

Fixes #896

### DIFF
--- a/criteria/make-the-codebase-findable.md
+++ b/criteria/make-the-codebase-findable.md
@@ -16,24 +16,22 @@ Well-written metadata containing a unique and persistent identifier, such as a W
 
 ## Requirements
 
-* The codebase MUST be findable using a search engine by describing the problem it solves in natural language.
-* The codebase MUST be findable using a search engine by codebase name.
 * The name of the codebase SHOULD be descriptive and free from acronyms, abbreviations, puns or organizational branding.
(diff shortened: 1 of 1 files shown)


ElenaFdR:
I think this should be 'should', not 'optional', since it would be better for codebase communities to be aware of how findable their codebase is via search engines so that they can take steps to improve findability.

Ainali:
I see your point, but I think that if we make it a SHOULD, then the contribution does not really solve the issue.  I think this because we didn't come up with any new or better ideas about how to evaluate whether the requirement is met or not.

ericherman:
I think `OPTIONAL`  for searching for the codebase by name, as that something that we can verify -- even if it involves paging through a long list of results.

ericherman (approved):

PR #907, opened 2023-05-01
docs/review-template.hml replaced .md

ericherman:
See 0bdbeae472d18aa1eb342d45c30998cc37c03b91


Ainali:
Good catch!

One of these changes are already in #904, do we want to merge that first?

Ainali (approved):

PR #910, opened 2023-05-02
Elevate status of natural language description

### LINKED ISSUE #899: Add requirement for a short high-level description of the codebase
In [the last community call](https://blog.publiccode.net/community%20call/2023/04/11/notes-from-community-call-6-april-2023.html) when talking about findability of a codebase, it was brought up that having a good short description of it is important. It should be short enough to be possible to use in the repository description field and in the bio of any social media handles for the codebase. I suggest going for a SHOULD requirement.

Ainali:
Add requirement for short description.
Elevate importance of natural language search (short description). Fixed bad grammar of name search requirement.

This works builds upon c8f90d978d311cb6b4b8389c3c8dad51cdc2ccbd

Fixes #899

### DIFF
--- a/criteria/make-the-codebase-findable.md
+++ b/criteria/make-the-codebase-findable.md
@@ -17,21 +17,24 @@ Well-written metadata containing a unique and persistent identifier, such as a W
 ## Requirements
 
 * The name of the codebase SHOULD be descriptive and free from acronyms, abbreviations, puns or organizational branding.
+* The codebase SHOULD have a short description that helps someone understand what the codebase is for or what it does.
 * Maintainers SHOULD submit the codebase to relevant software catalogs.
 * The codebase SHOULD have a website which describes the problem the codebase solves using the preferred jargon of different potential users of the codebase (including technologists, policy experts and managers).
-* It is SHOULD for the codebase to be findable using a search engine by codebase name.
+* The codebase SHOULD be findable using a search engine by codebase name.
+* The codebase SHOULD be findable using a search engine by describing the problem it solves in natural language.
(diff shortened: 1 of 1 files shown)


Ainali:
Ping @jgroenen: this is the result from our [last community call](https://blog.publiccode.net/community%20call/2023/04/11/notes-from-community-call-6-april-2023.html). Do you think it captures it well?

ericherman:
This is the combined diff:

https://github.com/publiccodenet/standard/compare/main...short-description#diff-9d35233d3d9f30f432084711016fe842a89e99ec6dd6ab336edee072585bd04e

jgroenen:
Hi, yes, I think so, although I thought that where is stated "should be findable by a search engine by XYZ" we talked about replacing that with "should document XYZ" as is now in the diff, combined with "this information should be somewhere where it can be indexed by a search engine", which is not in the diff, but maybe somewhere else..

Ainali:
> Hi, yes, I think so, although I thought that where is stated "should be findable by a search engine by XYZ" we talked about replacing that with "should document XYZ" as is now in the diff, combined with "this information should be somewhere where it can be indexed by a search engine", which is not in the diff, but maybe somewhere else..

Good point because this is perhaps not very obvious, but the important part for this now lies in the "how to test" section. There, we essentially say that we pass or fail this requirement by testing with the short description, which means that you have to document XYZ.

ericherman (approved):

ericherman:
Catalogues and search engines are different target, both important. When we test, the maintainers are welcome to point us to specific catalogues that we should check (although we may suggest additional ones as well). For search engines, the test of indexable is whether it can be found on any of the major search engines -- even several pages of results down the list.

ericherman:
Would other verification be better?

PR #914, opened 2023-05-09
Fix grammar of search engine requirement

Ainali:
The error happened when the optional turned into a should and only the level changed.


Ainali:
Oh, this could be closed as it is fixed by #910

PR #916, opened 2023-05-09
Allow workflow trigger by hand

ericherman:
I *think* this is the correct syntax to allow both by-hand triggering and trigger on a schedule.


Ainali (approved):
Let's try it!

Issue #949, opened 2023-06-12
[editorial] Readers Guide


fliepeltje:
Suggestion regarding a revised/edited version of the readers guide (refers to #938 )

I think it is best if you look at the PR for the suggested changes and then discuss them through the review function and we can refine it as we go.

PR #950, opened 2023-06-12
Editorial/readers guide

fliepeltje:
See issue #949

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -12,7 +12,6 @@ Codebase stewards review the roadmap monthly as part of our [backlog pruning ses
 
 ## Current work
 
-* Work through the issue backlog
 * As far as possible, make the standard Standard-compliant
 
 ## Near term
@@ -28,6 +27,7 @@ Codebase stewards review the roadmap monthly as part of our [backlog pruning ses
 * Add to README and index.md information (and eventually statistics) on adoption and use (issue #438)
 * Physical paper checklist
 * Becoming validated by multiple codebases
+* Release version 1.0.0 after a few codebases are compliant
 * Linkable requirements
 * New cover art
 * Register for ISBN
--- a/readers-guide.md
+++ b/readers-guide.md
@@ -4,58 +4,35 @@
 ---
 # Readers guide
 
-The Standard describes a number of criteria.
-All criteria have consistent sections that make it clear how to create great public code.
+The Standard describes criteria with consistent sections that make it clear how to create great public code.
 
-References to "policy makers", "managers", and "developers and designers" apply to anyone performing duties associated with these roles, regardless of their job title.
+References to **policy makers**, **managers**, and **developers and designers** apply to anyone performing duties associated with these roles, regardless of their job title.
 It is common for individuals to have duties which span multiple roles.
 
-Below is a brief explanation of each of these sections and how they are used within the criteria of the Standard.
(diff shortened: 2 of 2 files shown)


ericherman:
Hello @fliepeltje, the idea to convert headings to bullet points is one worthy of consideration. I can see a correctness angle that the items are perhaps not sub-sections themselves. I also see that it changes the presentation such that it no longer carries the visual theme of the criteria layout. I'll raise this to the other foundation staff and collect opinions.

ElenaFdR:
Hi @fliepeltje, thanks for thinking with us on the layout and clarity. 

What was your logic for changing the content under each subheading? We tried to make the overall tone quite friendly, and the role explanations respectful and easy to understand by mentioning example responsibilties. It looks like you've changed this to listing areas of responsibility and cut out all the chattiness?

I don't think that changing subheadings to bullets increases usability. This is because:

- nested bullets are hard to read; we usually try to rewrite content to avoid them
- on a full page of bullets, it's often harder to keep track of the context of bullets further away from the introduction
- changing the sections being defined from subheadings to bolded text behind a bullet makes it harder for someone to skim to find the right section if they're looking for specific information. Screenreaders also let people skip from subheading to subheading, while treating bolded text quite variably (with the result that accessibility guidelines recommend against using bold for emphasis)

I think the correct comparison for this page's structure is the glossary, where a term or concept is defined, rather than a long list.

fliepeltje:
So as someone who wants to implement this, I am thinking very much from the perspective of someone who is reading this document with the idea to extract the relevant information and apply it. In that role, I expect this to be more of technical manual and in that regard I value briefness over chattiness which is the core motivation to reduce the amount of content. 

Personally, I find subheadings can be distracting, but I am aware that this is subjective. The reason I dislike them is that they prolong the file just in terms of height. It's like spreading your content over 2 A4s when it can fit on a post-it note. As someone who just wants to know what to do here, I want the information quickly. 

I should note, this type of structure to me only makes sense for this chapter; it is not something I would like to extend to other documents too. I indeed like the ability to send hrefs to specific subsections to colleagues and from the point of view of accessibility, it is nicer and it is a structure I would keep. But for this section it makes no sense to me. 

As a more concrete example, consider this section:

> ## Developers and designers: what you need to do
>
> This section tries to specifically speak to developers and designers by offering them concrete actions they can perform in  their role.
> 
> Developers are usually more technically aligned and have more impact on the delivery of services than the previous groups.

The paragraphs below add very little information, but take up a lot of space and are not as easy to scan as some bullet points. 

I guess in the bottom line I'd just like the standard to be a bit shorter and to-the-point in terms of writing really focusing very hard on whether the content adds information or is more filler text. At the same time, I think some levity in a document is nice too, but I would rather see that though examples and anecdotes, but not mixed with the technical specifications.

ericherman:
Given that the current layout targets the [pdf version](https://github.com/publiccodenet/standard/releases/download/0.7.0/standard-for-public-code-0.7.0.pdf) we are likely to keep that format.

The text could be tightened up, if it is consistent with that format. Some of the text-shortening in this patch is achieved by converting full sentences to shorter bullet phrases, which would not be consistent with the format. That said, making the sentences shorter and sharper would be welcome.

fliepeltje:
Alright; I'll see if I can feed some more contributions like that when I find the time for it :)

PR #954, opened 2023-06-20
move version badge into repo

### LINKED ISSUE #948: version "badge" in README not in assets/
Low priority, as this is not in the printed version, only in the README.

We see the version shield sometimes being slow to load and failing somewhat regularly in our external link check.

Rather than rely upon an external service, we could generate that `.svg` ourselves, as we do the link-check badge, and check it in to `assets/`.

It would add a step to the [release procedure](https://github.com/publiccodenet/standard/blob/develop/docs/releasing.md), but not a big one.

We doubt that shields.io is going to go away any time soon, yet it is an external dependency that we no longer need since we are doing at least some badge generation ourselves.

ericherman:
fixes #948

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ urls-sorted.txt
 urls.txt
 release.body
 *.orig
+node_modules/
+package-lock.json
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 The Standard for Public Code gives public organizations a model for preparing open source solutions to enable collaborations with similar public organizations in other places.
 It includes guidance for policy makers, city administrators, developers and vendors.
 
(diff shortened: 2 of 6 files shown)


Ainali:
When trying to run the script I get 
```
./script/make-version-badge.sh: rad 41: ./node_modules/.bin/badge: Filen eller katalogen finns inte
```
(./script/make-version-badge.sh: line 41: /node_modules/.bin/badge: File or directory does not exist)

ericherman:
That's interesting, maybe it didn't run the "node install"? 

Please run it like "VERBOSE=1 script/make-version-badge.sh" and let's see the detail output

Ainali (approved):

PR #955, opened 2023-06-21
Move subsequent OMG format links to Further reading

### LINKED ISSUE #944: Link density of Bundle policy and source code
As of 0.7.0, we have really high link density in one of the bullets of "Public policy makers: what you need to do":

```
* Document policy in formats that are unambiguous and machine-readable such as [Business Process Model and Notation](https://en.wikipedia.org/wiki/Business_Process_Model_and_Notation), [Decision Model and Notation](https://en.wikipedia.org/wiki/Decision_Model_and_Notation) and [Case Management Model Notation](https://en.wikipedia.org/wiki/CMMN).
```

Should we split this, remove a link, or something else?

ericherman:
fixes #944

In the future, when we have concrete examples of codebases with policy documented in unambiguous and machine-readable formats, we can return to the "such as this or that", but as BPMN, DMN, and CMMN are all Object Management Group specifications, one should be enough. We can keep the others at the bottom, for now, but eventually I expect these to move out to the implementation guide.


Ainali:
Nice, but shouldn't we move all three down then and refer only to OMG in the requirement?

Ainali (approved):

Issue #942, opened 2023-05-31
Have release-on-tag workflow set the date of the release


Ainali:
When we write the CHANGELOG, it is not certain what date the release will be made. We should investigate if the workflow could set it when it runs.

ericherman:
This idea is good, and will require adjusting the release process as it raises questions about tagging.

The workflow would need to not only build from a tag that was pushed, but also push the changed `CHANGELOG.md` back as well.

Since we want the artefacts to match a specific the tag, perhaps the `0.9.1` build should make changes, and tag the new changes as `0.9.1-release` or similar?

Or, we could go the other way around: we could trigger release builds only if the tag is of a form that includes '-release' in the name, like `0.9.2-release`, and then have the `release-on-tag` workflow create the plain `0.9.2` tag after it has updated the `CHANGELOG.md`.

While I've never used a GitHub workflow to create new commits and tags and push them back, I am confident we can find a workflow example to work from.

PR #956, opened 2023-06-21
Sketch a script to update CHANGELOG date

ericherman:
This sketch of a script only updates the date in the CHANGELOG.

Rather than search and replace a date, better would be to put in an easier-to-replace marker e.g. SET_THIS_DATE

The larger question of how to change the release process and the workflow can be seen GitHub issue #942:

"Have release-on-tag workflow set the date of the release" https://github.com/publiccodenet/standard/issues/942


Ainali (approved):
While a marker might be better, this seems to work quite fine and is an improvement.

ericherman:
Superseded by https://github.com/publiccodenet/standard/pull/965

PR #957, opened 2023-06-21
Expand how to test codebase name

### LINKED ISSUE #822: "Codebase name should be free of acronyms" lacks nuance
For instance, a codebase with a name like `svg-toolkit` would make sense as `.svg` is well known and what people would search for if looking for Scalable Vector Graphics file format tools.

How can we express that acronyms or abbreviations which are (nearly) universally known can be okay in some situations?

ericherman:
fixes #822

### DIFF
--- a/criteria/make-the-codebase-findable.md
+++ b/criteria/make-the-codebase-findable.md
@@ -29,7 +29,9 @@ Well-written metadata containing a unique and persistent identifier, such as a W
 
 ## How to test
 
-* Check that the codebase name is free of acronyms, abbreviations, puns or organizational branding.
+* Check that the codebase name is descriptive and free of puns.
+* Check that the codebase name is free of acronyms and abbreviations or that the acronyms or abbreviations in the name are more universally known than the longer forms.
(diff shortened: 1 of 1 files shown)


Ainali (dismissed):
I like this approach a lot.

criteria/make-the-codebase-findable.md
```diff
@@ -29,7 +29,9 @@ Well-written metadata containing a unique and persistent identifier, such as a W
 
 ## How to test
 
-* Check that the codebase name is free of acronyms, abbreviations, puns or organizational branding.
+* Check that the codebase name is descriptive and free of puns.
+# Check that the codebase name is free of acronyms and abbreviations or that the acronyms or abbreviations in the name are more universally known than the longer forms.
```
Ainali, line 33:
```suggestion
* Check that the codebase name is free of acronyms and abbreviations or that the acronyms or abbreviations in the name are more universally known than the longer forms.
```

Ainali (commented):
The approach is good, but the syntax needs a tiny fix.

Ainali (approved):

Ainali (approved):

PR #961, opened 2023-06-27
Clarify use of "source code" vs "policy"

### LINKED ISSUE #208: Feedback: Glossary definition Code broader than use in document
As fed back by @EdoPlantinga

### LINKED ISSUE #496: Clarify when "Code" means "Source Code" or "Policy Text"
Feedback from policy makers is that the Standard should be clearer that source code and policy texts are two different things and use different words for them.

### LINKED ISSUE #958: Software is installed, deployed, or run ... not source code
For example, in criteria/document-the-code.md

* The documentation of the codebase SHOULD contain a section describing how to install and run a standalone version of the source code, including, if necessary, a test dataset.

Clearly this doesn't make sense for languages like Kotlin, Java, C, C++, golang, rust, etc.

And while this seems natural when thinking about interpreted languages like PHP/Perl/Python/Ruby, even with these the software, the running program, is the combination of the interpreter and the source code.

ericherman:
Fixes #208
Fixes #496
Fixes #958

### DIFF
--- a/criteria/bundle-policy-and-source-code.md
+++ b/criteria/bundle-policy-and-source-code.md
@@ -7,7 +7,7 @@ redirect_from:
 ---
 # Bundle policy and source code
 
-Access to both source [code](../glossary.md#code) and [policy](../glossary.md#policy) documentation provides building blocks for anyone to implement the codebase in their local context or contribute to the further development of the [codebase](../glossary.md#codebase).
+Access to both [source code](../glossary.md#source-code) and [policy](../glossary.md#policy) documentation provides building blocks for anyone to implement the codebase in their local context or contribute to the further development of the [codebase](../glossary.md#codebase).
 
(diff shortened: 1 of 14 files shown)


ericherman:
@Ainali perhaps before merging it, if you feel it fixes any of these issues, we could update the text of the PR comment to reflect that.

Ainali:
I think almost all the changes are good and that they will fix all three of the issues. 
There are two places where i want to keep the original wording with only code instead of changing it to source code. It's on line 11 in criteria/make-the-codebase-reusable-and-portable.md and line 21 in criteria/use-plain-english.md. 
In both places, I think the sentences make more sense when referring to the wider concept.

As a separate note, if we *do* change them both, we can also remove the definition of code from the glossary as we don't refer to it anywhere.

ericherman:
@Ainali please re-review

Ainali (approved):

PR #1044, opened 2024-10-12
Improve install instructions

ericherman:
Without specifying a bundle configuration to be local, bundle requires root and may conflict with the system.


Ainali (approved):

bvhme (approved):

janderssonse:
@ericherman I tested on Ubuntu 24.x from a pretty clean install I had, and I also had to install the package ruby-dev. So I suggest adding that to the apt line of the install instruction, and then it should be all set.

ericherman:
Thank you, @janderssonse I've added `ruby-dev` as well.

clausmullie (approved):

janderssonse (approved):

ericherman:
As an aside, I am a _little_ surprised that on Ubuntu `bundler` does not depend upon ruby-dev because many gems require it. By contrast, on my Debian box I see that it does:

```
$ apt-cache rdepends ruby-dev
ruby-dev
Reverse Depends:
  diaspora-installer
  ruby-build
  bundler
  ruby-full
  ruby
```

Regardless, it doesn't hurt to specify it explicitly, perhaps it is even clearer to do so.

ericherman:
rebased

janderssonse:
Ready to merge imho, it has 4 approvals:)

Issue #1058, opened 2024-12-02
Can I pull straight to main for the jekyll theme change? 


clausmullie:
Pulling develop into main triggers a release and version change on the standard. 

However, I'd like to switch to the community theme without triggering a release. Is it ok to do this by directly making a pull request on main given that it does not change anything on the standard itself?

(A release might be at order soon - separate issue at #1057)

ericherman:
> Pulling develop into main triggers a release and version change on the standard.

A merge to main will not trigger the workflow.

If we look at lines 4-7 of [`.github/workflows/release-on-tag.yml`](../blob/develop/.github/workflows/release-on-tag.yml#L4-L8) we can see that an automated release is only triggered if someone pushes a tag named with the form `trigger-1.2.3`.

Thus, if you wish, to re-deploy the website with the same contents as the current release, only changing the look and feel of the website, then you should be able to create a "hot-fix" branch based upon `main` (rather than `develop`) which changes the theme, and you can merge that branch to main without creating a new release.

Once you have done that, I suggest creating a merge request from `main` to `develop` so that develop is up-to-date with the new `main`.

If you would like a second set of eyes, I'd be happy to help you do this.

clausmullie:
O wait, it's not needed - I just realised github pages builds from the develop branch, so all the changes are live already.

PR #1059, opened 2024-12-03
updates to community theme

clausmullie:
Updating to community theme straight to the main branch (see #1058)

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -7,7 +7,10 @@ plugins:
   - jekyll-redirect-from
   - jekyll-sitemap
 
-remote_theme: publiccodenet/jekyll-theme
+remote_theme: standard-for-public-code/jekyll-theme
+
+#baseurl: /standard-for-public-code
+
 include:
   - "README.md"
   - "CONTRIBUTING.md"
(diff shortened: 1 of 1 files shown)


clausmullie:
Build locally, fails here. 

I see we updated to actions/checkout@v4 with 0ab837060a60e4eefe99e6104eec7d5a3155ddcd and more importantly, in `.github/workflows/test.yml` updated ruby version from 2.7 -> 3.1 with 9dca31f41cb5b7b723c54b0ad59f05b62ce46d37

I'm assuming it's this fix that would need to be implemented?

clausmullie:
Theme already live as github pages builds from the develop branch.

ericherman:
> Theme already live as github pages builds from the develop branch.

My memory might be wrong, but I believe this reflects a difference in the configurations of the repositories.

Regardless, it is worth considering if this is the desired behaviour, as the threshold for merging to develop is lower and likely far less rigorous than merging to main or making a release.
I suggest that this should be either an issue in the repository or perhaps a discussion topic in the next monthly call; I don't know which is best.

Ainali:
> I suggest that this should be either an issue in the repository or perhaps a discussion topic in the next monthly call; I don't know which is best.

Agreed, a conscious decision would be great.

PR #1060, opened 2024-12-20
Add further reading: developmentStatus in publiccode.yml

bvhme:
In order to provide a standardised model for documenting codebase maturity the `publiccode.yml` standard has maturity levels built in.

Standardising on these levels seems like a good idea.

Alternatively these could be added as a 'for example' in the criteria as well.

### DIFF
--- a/criteria/document-codebase-maturity.md
+++ b/criteria/document-codebase-maturity.md
@@ -46,5 +46,6 @@ Understanding how a codebase has evolved is key to understanding the codebase an
 
 * [Semantic Versioning Specification](https://semver.org/) used by many codebases to label versions.
 * [Software release life cycle](https://en.wikipedia.org/wiki/Software_release_life_cycle)
+* You can use the [Development Status key in the `publiccode.yml` standard](https://yml.publiccode.tools/schema.core.html).
 * [Service Design and Delivery Process](https://www.dta.gov.au/help-and-advice/build-and-improve-services/service-design-and-delivery-process) by the Australian Digital Transformation Agency.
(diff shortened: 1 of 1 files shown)


Ainali:
I like the resource as such! However, this would be the first and only instruction used in the Further reading section.

I would prefer we stick to the established form of: [title (hyperlinked)] by [source].

ericherman:
I would like to point to the `publiccode.yml` specification documentation of the maturity levels, but I do not see where this is documented .... am I missing something obvious?

Ainali:
> I would like to point to the `publiccode.yml` specification documentation of the maturity levels, but I do not see where this is documented .... am I missing something obvious?

They don't call it maturity levels, that may be the confusing thing. But as the text for the link refers to development status, I assume it is https://yml.publiccode.tools/schema.core.html#key-developmentstatus that @bvhme referred to.

bvhme:
I've updated it with the specific link. Using the format with the source seems a bit weird here.

The source is not that important here and could use some context. This makes me wondering if this should really be in the 'what to do' section or something.

Ainali:
Yeah, if it doesn't fit if written on that form, it is probably not further reading.

I am starting to think that this should go in the [community-implementation-guide-standard](https://github.com/standard-for-public-code/community-implementation-guide-standard) rather than in the Standard itself.

ericherman:
To me, the text "can be used to show if the codebase is ready to use" strikes me as both a bit out of place and somewhat stating the obvious given the context. What if we simplify it to simply be the link? Perhaps like:

*  [Development Status](https://yml.publiccode.tools/schema.core.html#key-developmentstatus) key of the publiccode.yml standard.

ericherman (approved):

Issue #1067, opened 2025-01-29
Community call agenda link on the web page is stale


ericherman:
index.md: "The agenda" [links](https://github.com/standard-for-public-code/standard-for-public-code/blob/develop/index.md?plain=1#L39) to an old publiccode dot net eitherpad

clausmullie:
Closed with [#1068](https://github.com/standard-for-public-code/standard-for-public-code/pull/1068)

PR #1068, opened 2025-01-29
Adds discussion board as agenda 

clausmullie:
Proposes fix for #1067

### DIFF
--- a/index.md
+++ b/index.md
@@ -36,7 +36,7 @@ Additional context and background can be found in the [foreword](foreword.md).
 ## Community calls
 
 We usually have a community call on the first Thursday of the month at 15:00 (CET/CEST).
-The agenda is coordinated on our [discussion board](https://github.com/standard-for-public-code/standard-for-public-code/discussions) roughly a week before the call.
+The agenda is coordinated on our [discussion board](https://github.com/standard-for-public-code/standard-for-public-code/discussions/categories/community-calls) roughly a week before the call.
 
 ## Other resources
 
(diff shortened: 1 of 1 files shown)


ericherman (approved):
I think this is good.

If we get so many discussions that it becomes hard to find in the future, we can revisit this decision.

index.md
```diff
@@ -36,7 +36,7 @@ Additional context and background can be found in the [foreword](foreword.md).
 ## Community calls
 
 We usually have a community call on the first Thursday of the month at 15:00 (CET/CEST).
-[The agenda](https://write.publiccode.net/pads/Community-Call-Standard-for-Public-Code) is updated roughly a week before the call.
+The agenda is coordinated on our [discussion board](https://github.com/standard-for-public-code/standard-for-public-code/discussions) roughly a week before the call.
```
Ainali, line 39:
```suggestion
The agenda is coordinated on our [discussion board](https://github.com/standard-for-public-code/standard-for-public-code/discussions/categories/community-calls) roughly a week before the call.
```
Making the link directly to the community calls category should ensure that even if a lot of discussions are going on about other things, it will be easy to spot the agenda.

Ainali (changes requested):

clausmullie:
Good catch @Ainali !

ericherman (approved):
Ah, even better!

Ainali (approved):

janderssonse (approved):
LGTM!

PR #1091, opened 2025-04-07
Copyright 2025

### LINKED ISSUE #1088: Update ending year in SPDX header
The ending year of 2024 should be changed to 2025.

ericherman:
Note that the SPDX specification states that the cardinality of the copyright field is 0 or 1:

https://spdx.github.io/spdx-spec/v2.3/file-information/#88-copyright-text-field

As such, this uses semicolons to separate separate copyright lines.

fixes #1088


Ainali (commented):
Is the same AUTHORS link for the two parts intentional, or should the second go the Foundation?

ericherman:
Previously, the copyright was held by the combination of the Foundation for Public Code and the contributors in the AUTHORS file.  Of course, contributions from 2025 forward are only the contributors in the AUTHORS file. I believe the copyright text in this contribution represents that reasonably well. 

I think that the second link should _not_ go to the foundation site because the AUTHORS file is (and was) maintained in the main (upstream) repository.

Should any downstream version maintain a different AUTHORS file (rather than push upstream), then I'd also expect that they'd also have a slightly different copyright text line and different link, but probably not for any previous years.

That said, if we believe we should show a different AUTHORS link in order to show a smaller AUTHORS list prior to 2025, we could link to an earlier commit-hash of this AUTHORS file, but I don't view that as making a ton of sense to do now because that wasn't seen as a need in prior years.

Naturally, I'm happy to adjust the contribution to match the consensus if the prevailing view is that the copyright text should be different from what is in the contribution as it stands now.

ixuz (approved):
Looks good

PR #1100, opened 2025-04-11
Updated review template for release 0.8.1

ixuz:


docs/review-template.html
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en" dir="ltr">
 <!-- SPDX-License-Identifier: CC0-1.0 -->
-<!-- SPDX-FileCopyrightText: 2022-2024 The Foundation for Public Code <info@publiccode.net>, https://standard.publiccode.net/AUTHORS -->
+<!-- SPDX-FileCopyrightText: 2022-2025 by The Foundation for Public Code <info@publiccode.net>, https://standard.publiccode.net/AUTHORS -->
```
ericherman, line 4: I wouldn't expect a "by" here

ericherman (changes requested):

docs/review-template.html
```diff
@@ -45,11 +45,11 @@
 </style>
 </head>
 <body>
-<h1>________ and the Standard for Public Code version <span class="standard-version">0.8.0</span></h1>
+<h1>________ and the Standard for Public Code version <span class="standard-version">ba42e7f</span></h1>
```
ericherman, line 48: As this was done by hand, the version should probably be empty for this commit, as this version is not yet tied to any tag. The automation will regenerate it with the tag's version when it runs.

We could change the automation to detect the by-hand/no-tag case.

ericherman (approved):

Issue #1062, opened 2024-12-25
gh-pages release process review


ericherman:
As discussed in #1059, it seems that the repository is configured to build the `gp-pages` website based on the `develop` branch rather than the `main` branch as it was previously.

The maintainers should review with the community whether or not this is desired, and either update the configuration or the documentation accordingly.

ericherman:
From: Minutes Jan. 29, 2025 #1070 

"Decision: Shift the GitHub pages deploy branch to main and use regular releases. It will be done after the next release, as not to revert changes in theme."

ericherman:
The docs/releasing.md has the steps, but lacks an introductory paragraph with an overview.

ericherman:
The docs/releasing.md step 4 needs to be expanded to be clear about the release candidate process; suggest splitting into two steps, one for the -rc1 (and -rc2 if needed) and another for the actual non-rc tagging.

ixuz:
I've adjusted the gh-pages to deploy from `main`-branch.

PR #1109, opened 2025-04-20
fixes community call date from first to last Thursday of the month

clausmullie:
Propose a hot fix to main for #1108

Which we can then merge into the release branch and develop


ericherman:
If it should be deployed before 28 April, then hotfix on main makes sense.

If that's the case, then after merge, please open a PR from main to release-0.8.1 ... so that it can be merged to the release branch before the release is triggered.

And I suggest to _not_ create a PR from main to develop, as that merge will be [done as part of the release automation](https://github.com/standard-for-public-code/standard-for-public-code/blob/develop/.github/workflows/release-on-tag.yml#L110-L112) on 28 April, and would create a more complicated merge history.

ericherman:
> If it should be deployed before 28 April, then hotfix on main makes sense.

No wait!  hotfix to `main` does not make sense, because the website is currently deployed from `develop` not `main`.

Therefore, this could be done in `develop` and merged to the release branch, but not `main`.

After this release, then [deploy will be switched to happen from `main`](https://github.com/standard-for-public-code/standard-for-public-code/issues/1062#issuecomment-2634936793)

ericherman (changes requested):
Change the target branch from `main` to either `release-0.8.1` if it can wait until 28 April, or `develop` if it should be deployed immediately.

clausmullie:
> No wait! hotfix to `main` does not make sense, because the website is currently deployed from `develop` not `main`.

O right! Yes. I don't see a reason to wait until 28th, as it's an error and the next call is on the 24th.

I'll do in into develop.

ericherman (changes requested):
The branch this change is based upon is still main, even though it is trying to make a single change to develop.

The conflict with the CNAME and is likely to conflict again with [PR #1105](https://github.com/standard-for-public-code/standard-for-public-code/pull/1105).

Please rebase the branch on develop, so the PR has only the single commit in it.

ixuz (changes requested):
Indeed, rebase accordingly to @ericherman recommendation

ixuz:
I rebased this hotfix onto `develop`. It should be good now.

ixuz (approved):

ericherman (approved):

clausmullie:
Great, thanks @ericherman and @ixuz !

Issue #1101, opened 2025-04-14
Describe the process of template version diff to update an assessment


ixuz:
Please have a look at this @ericherman and @Ainali

Ainali:
The description is available here: https://github.com/publiccodenet/about/blob/eb2ee65b76f78e1a1853c44e2d1ef0982240541d/activities/releasing-the-standard/assessment-update.md

PR #1132, opened 2025-04-30
Update assessment diff patch

ericherman:
In #1101 @Ainali found the original description of the process.

I used `git format-patch` to export that description, then adjusted the resulting patch slightly to put it in the `docs/` directory, and then used `git am` to import the patch.


ericherman:
@Ainali @ixuz I believe this is ready for review

docs/assessment-update.md
```diff
@@ -0,0 +1,47 @@
...
+--- a/path/to/standard-for-public-code-assessment.html
++++ b/path/to/standard-for-public-code-assessment.html
+```
+
+## Apply the patch
+
+From a clone of the assessed repository, apply the modified patch:
+
+```bash
+patch -Np1 -i /path/to/modified-review-template.patch
```
ixuz: Shall we avoid usage of shorthand flags for this command?

```bash
patch -Np1 -i /path/to/modified-review-template.patch
```

`patch --help`

`-N  --forward  Ignore patches that appear to be reversed or already applied.`
`-p NUM  --strip=NUM  Strip NUM leading components from file names.`
`-i PATCHFILE  --input=PATCHFILE  Read patch from PATCHFILE instead of stdin.`

Would this command be equivalent?

```bash
patch --forward -strip=1 --input=/path/to/modified-review-template.patch
```

ericherman: Yes, very good idea. I shall adjust it for long-options. It might be a couple of days.

ixuz (approved):
Looks good

Ainali (approved):
This is great!

I think we can make the step before a bit clearer too now, but that should not block this improvement, I'll make a new issue for my idea.

ixuz:
Alright, I'll take that as this change is approved and good to go.

PR #1135, opened 2025-05-08
Explicitly declare permissions of workflows to adhere to the principle of least privilege

ixuz:
GitHub security code scanning reported the lack of explicit permissions for these workflows.

### DIFF
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -2,6 +2,10 @@
 # SPDX-FileCopyrightText: 2025 Standard for Public Code Authors, https://www.standardforpubliccode.org/AUTHORS; 2021-2024 The Foundation for Public Code <info@publiccode.net>, https://www.standardforpubliccode.org/AUTHORS
 
 name: local link check
+
+permissions:
+  contents: read
+
 on:
   workflow_dispatch: {}
 #  schedule:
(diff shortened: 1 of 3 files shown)


ixuz:
I believe I've set the right permissions for each workflow. Somebody please confirm.

ericherman:
I can confirm that you've correctly identified which workflows are read-only and which one requires write access.

Having not used this syntax before, I can not confirm that they are correct or even do what is expected; I guess we have to take that on faith.

ixuz:
> I can confirm that you've correctly identified which workflows are read-only and which one requires write access.
> 
> Having not used this syntax before, I can not confirm that they are correct or even do what is expected; I guess we have to take that on faith.

So.. shall we just send it and hope it's ok? 🚀😅
(I need an approval before doing that though)

ericherman:
> > I can confirm that you've correctly identified which workflows are read-only and which one requires write access.
> > Having not used this syntax before, I can not confirm that they are correct or even do what is expected; I guess we have to take that on faith.
> 
> So.. shall we just send it and hope it's ok? 🚀😅 (I need an approval before doing that though)

I'm not an approver, however if I were, I'd suggest that there's only one way to find out: "let 'er rip!"

ericherman (approved):
This looks good to me as far as I can tell.

I have no way to test this except to try it.

zorp (approved):

Issue #1134, opened 2025-05-01
Stronger guidance for contributors to add themselves to AUTHORS


ericherman:
From [1094#issuecomment](https://github.com/standard-for-public-code/standard-for-public-code/issues/1094#issuecomment-2796355329) :

Stronger guidance should be provided in the [CONTRIBUTING](https://github.com/standard-for-public-code/standard-for-public-code/blob/develop/CONTRIBUTING.md#4-celebrate) file.

Today, it simply reads:

> In fact, feel free to open a pull request to add your name to the [`AUTHORS`](https://github.com/standard-for-public-code/standard-for-public-code/blob/develop/AUTHORS.md) file and get eternal attribution.

That could be a stronger request, along the lines of "If your name is not already in AUTHORS, please open a pull request to add yourself."

(This does not remove the need to manually check AUTHORS during release, as new contributors may not have been added yet for any number of reasons.)

Ainali:
Fixed by #1137 (but I don't have permission to close the issue).

PR #1137, opened 2025-05-15
tweaks author instructions

clausmullie:
Proposes fix for #1134

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,7 +89,7 @@ If your documentation and code have passed human review, it is merged.
 Your ideas, documentation and code have become an integral part of this project.
 You are the open source hero we need!
 
-In fact, feel free to open a pull request to add your name to the [`AUTHORS`](AUTHORS.md) file and get eternal attribution.
+If your name is not already in [`AUTHORS`](AUTHORS.md), please feel free to open a pull request to add yourself to get eternal attribution 🎉.
 
 ## Languages and translations
 


ericherman:
Should it still have 'feel free'?

I think it should be more of a specific request of contributors along the lines of, "if you are not in the list, please add yourself".

For a request of others (which might be uncomfortable or represent more work than they would otherwise do), I favour boring-and-direct over trying to make it sound exciting, thus I'm "eyerolly" (🙄) about adding another emoji there, and suggest that be dropped.

This line of thinking also suggests to me that "correct attribution" would be more appropriate than "eternal attribution".

clausmullie:
Is it a requirement people add themselves? Or voluntary if they would like it? 

Perhaps someone might not want to be on the list? Or will add them regardless if they don't?

Ainali:
No, it's not a requirement. The purpose is, however, to make it easier during release where we hopefully don't have to add anyone. Come to think of it, it may be worth adding to our review instructions that if a contribution is substantial and the contributor is not in the authors yet, to suggest that they do it in that pull request or explicitly decline. That way, we would also know if they don't want to be in it and give encouragement to those who wouldn't be bold enough otherwise.

ericherman:
I'd say it's like a SHOULD requirement, not a MUST requirement; a tongue-in-cheek framing might be:

* The repository itself MUST have correct attribution
* The repository MUST NOT accept contributions which are not the contributor's right to contribute
* Contributors SHOULD add themselves to the author list
* It is OPTIONAL for contributors to not be listed in the published author list, provided that there is record of the contribution's authorship in the history

bvhme (approved):

clausmullie:
How about something along the lines of
```
As a community we would like to include all contributors as [`AUTHORS`](AUTHORS.md).
If your name is not already there, please open a pull request to add yourself.
Each release will include a review to check if all new contributors have been mentioned.
If you wish not to be included, please let us know explicitly.
```

ericherman:
I like this; it is very clear and explicit.

I note that historically, we never really asked anyone if they did _not_ want to be included in AUTHORS (and I expect this to be exceedingly rare), but we did often ask contributors how they would like to be listed, and thus we would have learned if someone did not wish to be listed at all. Thus, I do not expect many toes were stepped upon.

Ainali:
> How about something along the lines of

Yes! This is great because it is explicit, transparent and radiates intent.

Ainali (approved):

ericherman (approved):

PR #1146, opened 2025-06-26
Updates links to renamed implementation repo

clausmullie:
Fixing straight to main as links are currently broken with repo name change

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ To see how ready your codebase is, you can do a quick [eligibility self assessme
 
 The standard *should* be mostly self-explanatory in how to apply it to your codebase.
 If anything in the standard is unclear, we encourage you to open an issue here so that we can help you and anyone else who feels the same as you.
-For inspiration, look at the [community built implementation guide](https://standard-for-public-code.github.io/community-implementation-guide-standard/) which contains examples and other tips.
+For inspiration, look at the [implementation guide](https://standard-for-public-code.github.io/implementation-guide-standard/) which contains examples and other tips.
 
 If there are any breaking changes in a new version of the Standard for Public Code, the codebase stewards at the Foundation for Public Code will help any implementers of the standard understand how the gaps can be closed.
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -112,7 +112,7 @@ December 5th 2022: 📝 the twelfth draft clarifies document maturity.
 * Add remaining SPDX headers to files.
 * Made additional minor changes to text for clarity.
 * Some hyperlinks updated.
-* Moved examples to the [Community implementation guide](https://standard-for-public-code.github.io/community-implementation-guide-standard/).
+* Moved examples to the [Implementation guide](https://standard-for-public-code.github.io/implementation-guide-standard/).
 
 ## Version 0.4.0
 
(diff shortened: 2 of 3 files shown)


clausmullie:
Reverted the repo name change so there is no urgency. That can be done when this has been reviewed. Should also include a check on other repos to see if they point to the implementation guide

ixuz (approved):
I would've guessed that the URL of a moved GitHub repository will automatically become a 301 redirect to the new location, but I'm not 100% sure. What I'm saying is that perhaps we could move the implementation guide repository if we verify this behavior beforehand.

Keep in mind that the new links in this change, leads to a `404 - Not found` until the target repository has moved.

So I think the most graceful order of action would be to:

1. Move the repo (the original url _should_ become a redirect)
2. The merge this change.

Overall this change looks good.

Ainali:
> Keep in mind that the new links in this change, leads to a `404 - Not found` until the target repository has moved.

Just highlighting this! We should **not** merge this until that is fixed.

ixuz:
> I would've guessed that the URL of a moved GitHub repository will automatically become a 301 redirect to the new location, but I'm not 100% sure. What I'm saying is that perhaps we could move the implementation guide repository if we verify this behavior beforehand.
> 
> Keep in mind that the new links in this change, leads to a `404 - Not found` until the target repository has moved.
> 
> So I think the most graceful order of action would be to:
> 
> 1. Move the repo (the original url _should_ become a redirect)
> 2. The merge this change.
> 
> Overall this change looks good.

I've performed a small test of this behavior, and it does not properly redirect.

1. Created an example-repository, containing an index.html file, and activated github pages.
2. I confirmed that the github pages site could be accessed and I saw the index.html content.
3. Then I renamed the `example-repository` to `example-repository-moved`.
  - This caused the original url to the github pages to break with `404 - Not Found`

ElenaFdR-ICATT:
The old Foundation Jekyll setup included redirect functionality ([example page with redirect applied](https://github.com/publiccodenet/about/blob/eb2ee65b76f78e1a1853c44e2d1ef0982240541d/roles/intern.md?plain=1#L5)). Do we still use enough of their Jekyll to reuse this? Has the tech moved on/is there a better way to do this now?

ixuz:
> The old Foundation Jekyll setup included redirect functionality ([example page with redirect applied](https://github.com/publiccodenet/about/blob/eb2ee65b76f78e1a1853c44e2d1ef0982240541d/roles/intern.md?plain=1#L5)). Do we still use enough of their Jekyll to reuse this? Has the tech moved on/is there a better way to do this now?

Okay, I've had a brief look into this. My conclusion is that we will need to keep a repository with the name `community-implementation-guide-standard` that contain a Jekyll redirect instruction to the new address `/implementation-guide-standard`.

The required steps would likely be:

1. Rename the [community-implementation-guide-standard](https://github.com/standard-for-public-code/community-implementation-guide-standard) repository to `implementation-guide-standard`.
2. Create a new repository with the old repository name `community-implementation-guide-standard` with a Jekyll instruction redirect to `/implementation-guide-standard`.
3. Approve and merge [this PR #1146](https://github.com/standard-for-public-code/standard-for-public-code/pull/1146).

Does that seem right? And do we want to proceed with it this way?
@ElenaFdR-ICATT @ericherman

PR #1154, opened 2025-07-31
Fix typo of version number in title

Ainali:
Somehow the title wasn't updated when the rest of the document was.


ericherman (approved):

ericherman:
That line would not be part of a cleanly applied patch and thus would require manual editing.

janderssonse (approved):

janderssonse:
With two approves, and a very small changeset, I'll merge this.

